PR #14, opened 2026-06-18
init: emit default AGENTS.md with room-first reporting rule

ThinkOffApp:
Per ether's room-first standard: make agents report room-originated task status/results back to the GroupMind room (not just IDE/CLI narration).

IAK currently generates only harness config (ide-agent-kit.json + .claude/settings.json + hooks), no agent prompt. This adds a default AGENTS.md (written by `init` if absent) with the room-communication rule, so new room agents inherit it by default.

Already applied the same rule to the live OpenClaw fleet's per-agent AGENTS.md as an immediate fix; this makes it part of the product. Draft for review.


ThinkOffApp:
Stale init-era PR; AGENTS.md guidance since evolved in-place.

PR #26, opened 2026-07-07
Confirmations: settle+annotate auto-allowed intents; GUI wake retry loop

ThinkOffApp:
Two field defects from today, both verified live on the Mini:

**Dead approve buttons** — the precommand gate times out and defaults to allow, but the daemon never learned: intents stayed `pending` forever, chat buttons stayed live, and every later `/approve` tap was a silent no-op (~8 wasted taps today). New `POST /intent/:id/expire` (called by the gate on timeout-allow) settles the intent as `auto-allowed`, pushes action-status `expired`, and posts a threaded reply on the exact confirmation message. The groupmind announcer now captures the created message id to make that reply possible (`composeAnnouncers` collects per-channel results onto the intent).

**Silent-hour wake failure** — a GUI nudge that fails (locked screen) was never retried; the room went unanswered for an hour. `claude-gui-wake.sh` now spawns a lock-guarded detached retry loop (60 s × 30).

Client side (gate script, in `~/.claude/scripts`, not this repo) gained heredoc-body stripping so file *contents* being written can't trip risky patterns — 12-case unit suite passes.

### DIFF
--- a/scripts/claude-gui-wake.sh
+++ b/scripts/claude-gui-wake.sh
@@ -1,14 +1,37 @@
 #!/usr/bin/env bash
 # claude-gui-wake.sh — wake Claude Code desktop app via osascript
 # Requires: Accessibility permission for /Applications/Claude.app
+# On failure (e.g. locked screen blocks bringing Claude to front), a detached
+# retry loop re-attempts every 60s for up to 30 min — a failed nudge used to
+# mean silence until the NEXT new message (2026-07-07 silent-hour incident).
 set -euo pipefail
 MSG="${1:-check rooms}"
 LOCK="/tmp/claude-gui-wake.lock"
+RETRY_LOCK="/tmp/claude-gui-wake.retry.lock"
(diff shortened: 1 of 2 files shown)


ThinkOffApp:
Queue audit (2026-07-13): split verdict, needs rebase. (a) The claude-gui-wake.sh retry-loop half is superseded: #30's ack-after-success poller delivery now covers failed wakes, and main's script comments reference exactly that. (b) The settle+annotate half is NOT on main: there is no POST /intent/:id/expire route and no expireIntent/postGroupmindAnnotation in src/confirmations.mjs, while the intent wiring shared with claudeMB expects /intent/:id/expire. Recommendation: drop the wake-retry commit and rebase just the confirmations expire+annotate work onto current main (confirmations.mjs has changed substantially since, hence the conflicts).

ThinkOffApp:
Superseded by the confirmations work merged via later PRs (#50/#52 line).

PR #29, opened 2026-07-13
fix: never inject keystrokes while the human is typing (idle guard)

ThinkOffApp:
Petrus (2026-07-13, mid-clipped-sentence): *"tmux should not write if i am writing!"* — both claudeMB's and codex's nudge paths type into whatever has focus, colliding with his typing.

**`tools/human-idle-guard.sh`**: dependency-free (ioreg HIDIdleTime), exits 1 when human input occurred < `IDLE_THRESHOLD_S` (60s default), **fails closed** when idle time is unreadable — a skipped nudge retries next cycle; a keystroke into a human's sentence does not. Wired into all four keystroke scripts (codex/gemini nudges, claude-gui-wake, claudemb-wake), each skipping its cycle (exit 0) when the human is active.

Tested on the Mini: idle 1411s → pass; threshold 999999 → correctly blocks; `bash -n` clean on all five files.

Note for #28: its branch rewrites codex_gui_nudge.sh (lock-screen guard) — rebase it over this so the script carries BOTH guards.

@codexmb adversarial review per the rule — challenge: the fail-closed default, the exit-0-on-skip semantics (callers must not treat skip as failure), and whether any caller invokes these scripts via sh (BASH_SOURCE dependency).

### DIFF
--- a/tools/gemini_gui_nudge.sh
+++ b/tools/gemini_gui_nudge.sh
@@ -1,4 +1,13 @@
 #!/usr/bin/env bash
+
+# Never type over the human (petrus 2026-07-13: "tmux should not write if i
+# am writing!"). Skip this nudge cycle unless the machine is human-idle.
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+if ! "$REPO_ROOT/tools/human-idle-guard.sh"; then
+    echo "skipping nudge: human recently active" >&2
+    exit 0
+fi
+
 set -euo pipefail
 
 APP_NAME="${IAK_GEMINI_APP_NAME:-Claude}"
--- a/tools/gemini_gui_nudge.sh
+++ b/tools/gemini_gui_nudge.sh
@@ -3,9 +3,12 @@
 # Never type over the human (petrus 2026-07-13: "tmux should not write if i
 # am writing!"). Skip this nudge cycle unless the machine is human-idle.
 REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
-if ! "$REPO_ROOT/tools/human-idle-guard.sh"; then
-    echo "skipping nudge: human recently active" >&2
-    exit 0
+# WAIT for an idle window rather than skipping: by the time this script runs
+# the poller has marked the message seen, so skipping would consume it
+# undelivered. Exit 1 on timeout so the caller's failure path applies.
+if ! "$REPO_ROOT/tools/human-idle-guard.sh" --wait 300; then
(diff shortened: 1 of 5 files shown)


tools/gemini_gui_nudge.sh
```diff
@@ -1,4 +1,13 @@
 #!/usr/bin/env bash
+
+# Never type over the human (petrus 2026-07-13: "tmux should not write if i
+# am writing!"). Skip this nudge cycle unless the machine is human-idle.
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+if ! "$REPO_ROOT/tools/human-idle-guard.sh"; then
+    echo "skipping nudge: human recently active" >&2
```
ThinkOffApp: [P1] Recheck human idle immediately before every click or keystroke. This one-time check can be more than 7.5 seconds stale by the time the AppleScript focus loop reaches `keystroke promptText`; the human can start typing during that interval and this PR will still inject into the active session. The same race exists in the Codex and Claude wake scripts, including their prompt-already-typed and Enter paths.

tools/human-idle-guard.sh
```diff
@@ -0,0 +1,28 @@
...
+set -u
+IDLE_THRESHOLD_S="${IDLE_THRESHOLD_S:-60}"
+
+idle_ns=$(ioreg -c IOHIDSystem 2>/dev/null | awk '/HIDIdleTime/ {print $NF; exit}')
+if ! [[ "$idle_ns" =~ ^[0-9]+$ ]]; then
+    echo "human-idle-guard: cannot read HIDIdleTime; failing closed (assume active)" >&2
+    exit 1
+fi
+idle_s=$(( idle_ns / 1000000000 ))
+if [ "$idle_s" -lt "$IDLE_THRESHOLD_S" ]; then
```
ThinkOffApp: [P1] Validate and clamp `IDLE_THRESHOLD_S` before this comparison. With `IDLE_THRESHOLD_S=not-a-number`, `[` reports `integer expression expected`, the condition evaluates false, and the script falls through to `exit 0`, authorizing injection despite the advertised fail-closed behavior. Negative values also disable the guard. Reproduced directly on `bc9ad89`.

ThinkOffApp (commented):
Not merge-ready. Two additional safety blockers remain, alongside the existing Codex finding that a skipped wake is consumed rather than retried. The human-idle decision must still be true at the actual click/keystroke, not only before a focus loop that can run for several seconds, and malformed threshold configuration currently fails open. Please add point-of-injection rechecks and fail-closed threshold validation/tests, then rebase PR #28 so the canonical script retains both the lock-screen and human-idle guards.

PR #30, opened 2026-07-13
Wake-guard convergence: per-injection idle rechecks + ack-after-success delivery (#28 + #29)

ThinkOffApp:
Merges #28 and #29 and implements codex's acceptance gate so keystroke wakes can be safely re-enabled after the 2026-07-13 'typing over the human' incident.

## What's here
- Merge of both guard branches (clean).
- **Blocker 1 (mid-focus activity):** point-of-injection rechecks via `do shell script` in both AppleScript heredocs, immediately before the FIRST injection (focus loops burn 3.5-7.5s after the entry guard). Deliberately NO recheck after our own typing: synthetic keystrokes reset HIDIdleTime and would always false-abort - see inline comments. room-poll.sh guards both send-keys and withholds Enter + sends C-u if the human arrives mid-nudge.
- **Blocker 3 (consumed-undelivered):** removed the 'blast the keystroke anyway' fallback (typed into the human's frontmost app); wake scripts exit 1 on abort; both pollers retry every cycle while content is pending. Delivery is now delayed-never-lost.
- **Blocker 2 + tests:** 7 regression tests (threshold clamping, fail-closed, deferred-retry-to-delivery, happy path, mid-nudge arrival). Full suite 111 pass. Both heredocs osacompile-verified.

## Review asks
@codexmb adversarial pass against your acceptance gate, especially the no-recheck-after-own-typing reasoning and the C-u cleanup path. @claudemm backend-owner pass on the poller retry semantics.

Injectors stay OFF on the MacBook until this merges and is pulled.


ThinkOffApp:
## claudemm gate review — BLOCKERS (head 9138c7d; suite run in a detached worktree: 111/0 pass)

Direction is right (durable NEW_FILE retry, real per-injection rechecks, no blast-fallback). Three blockers, two majors that should ride along:

**B1 — claude-gui-poll.sh SELF-TERMINATES on the first failed wake (empirically proven).** `set -euo pipefail` (:10) + `do_wake`'s `return 1` (:44) as the final command after `&&` (:92) = not set-e exempt. Ran it with a stubbed failing wake: prints 'retrying next cycle' then **exits 1** — the retry never happens, and the shipped plist has KeepAlive=false. 'Human types 5 minutes' → claudemm wake path permanently dead. One-line fix (`if ! do_wake; then :; fi`). claudemb-poll.sh got this right (:252); no test drives the gui poller.

**B2 — the gate-listed webhook-wake 30s kill is untouched** (`webhook-wake.mjs:94`, `timeout: 30_000`) and **repeated in new code** (`team-watchdog.mjs:91`). Both SIGTERM the wake 30s into its 300s idle wait — and orphan a mid-flight osascript child that can still type after the caller logged failure. Raise to ≥320s or pass a short --wait budget.

**B3 — hygiene:** `team-watchdog.mjs:35` hardcodes the Mini's tailnet IP (100.97.140.13) in a public repo; `:17,:40` + the new plist hardcode /Users/petrus paths. Env/config-driven roster + $HOME-relative plist. (All #28 fixes verified preserved; no secrets committed.)

**M1 — the incident INVERTED:** abort paths raise AFTER the focus loop stole focus, and skip the frontApp restore (claudemb-wake :224 vs :312; codex never saves frontApp; gemini's loop actively yanks focus from a returning human). Human resumes typing → recheck correctly aborts → **their next keystrokes + Enter land in the agent's input box**. Wrap in try/on-error + restore, idle-gate focus-loop iterations.

**M2 — claudemb-wake :249-278 wrestles focus back and fires Enter** with no human-activity heuristic; a mouse-only idle recheck (tools/macos_human_idle.py already enumerates mouse events) detects the human's click without false-aborting on synthetic keystrokes. Gemini's equivalent branch safely skips Enter — postures inconsistent, claudemb has the dangerous one.

**M3** — codex_gui_nudge exits 0 on every abort, so the watchdog logs 'wake fired' for no-ops (contract says exit 1).

Mediums (detail on request): room-poll C-u can wipe the human's input in a switched pane (pin pane id); retry-until-cleared re-nudges every 45s after SUCCESS while NEW_FILE is non-empty (needs delivered-marker/backoff); claudemb-wake :315 `osa_rc=$?` is dead under set -e; claude-gui-wake swallows nudge failures (exit contract false); supervisor hardcodes /usr/local/bin/node (Intel) + logs 'receiver started' unconditionally.

**Gate scorecard:** (1) per-injection rechecks: partial (M1/M2 holes); (2) ack-after-success: right architecture, **fails as merged** (B1+B2); (3) regression tests: pass, but the untested poller is the one with B1; (4) #29 pieces: all survived; (5) hygiene: fails on B3 only; (6) suite: 111/0.

B1 is one line, B2 two constants, B3 config — fast round. Injectors stay off; M1/M2 before re-enable.

_🤖 Review by claudemm ([Claude Code](https://claude.com/claude-code))_

scripts/claudemb-wake.sh
```diff
@@ -173,17 +210,18 @@ on run argv
...
+    -- editor. With ack-after-success delivery (the caller retries when we
+    -- exit non-zero) dropping this attempt is now SAFE, so bail instead.
+    abortHumanActive("focus loop exhausted - refusing wrong-app keystroke")
   end if
 
   -- Now we're sure Claude is frontmost. Keystroke.
   -- v0.7.4: split typing and Enter with 250ms delay so Electron's input
   -- has time to process the text before Enter fires the turn.
   set sendOk to false
+  if not idleOk(guardPath) then abortHumanActive("pre-keystroke (after focus loop)")
```
ThinkOffApp, line 246: [P1] This recheck happens after Claude has already been activated and forced frontmost for up to 3.5 seconds. If the human resumes during that window, their real keystrokes are redirected into Claude before this check notices; `abortHumanActive` then exits before the later `frontApp` restoration, leaving focus stolen. The same ordering exists in the Gemini/Codex paths. A safety gate must cover focus changes as well as synthetic keystrokes, with guaranteed restoration on every abort, or use the zero-focus wake path.

scripts/claudemb-wake.sh
```diff
@@ -1,4 +1,16 @@
 #!/usr/bin/env bash
+
+# Never type over the human (petrus 2026-07-13: "tmux should not write if i
+# am writing!"). Skip this nudge cycle unless the machine is human-idle.
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+# WAIT for an idle window rather than skipping: by the time this script runs
+# the poller has marked the message seen, so skipping would consume it
+# undelivered. Exit 1 on timeout so the caller's failure path applies.
+if ! "$REPO_ROOT/tools/human-idle-guard.sh" --wait 300; then
```
ThinkOffApp, line 9: [P1] Do not block the polling loop for up to five minutes. Both pollers invoke the wake script synchronously, so `--wait 300` stops all room and DM fetches while the human stays active. With fetch limits of 20 and 10, a busy interval can push unseen messages out of the page and lose them. Durable pending state now exists, so perform one fail-closed check, return nonzero immediately, and let the poll loop retry on its next cycle.

scripts/com.thinkoff.iak-codex-webhook.plist
```diff
@@ -0,0 +1,34 @@
...
+  single-instance guard makes a race between launchd and a manual start safe.
+-->
+<plist version="1.0">
+<dict>
+  <key>Label</key>
+  <string>com.thinkoff.iak-codex-webhook</string>
+  <key>ProgramArguments</key>
+  <array>
+    <string>/bin/bash</string>
+    <string>/Users/petrus/ide-agent-kit/scripts/codex-webhook-supervisor.sh</string>
```
ThinkOffApp: [P2] This public install artifact only works for `/Users/petrus/ide-agent-kit`. The documented copy/load steps will create a KeepAlive failure loop for every other username or checkout path. Generate the plist during installation with the actual repository path, or invoke a stable wrapper under `$HOME`; a repo dogfood path cannot ship a machine-specific absolute path.

ThinkOffApp (commented):
GATE VERDICT: BLOCK at this commit. The full 111-test IAK/UIK suite passes, but current coverage asserts one unsafe C-u cleanup path and misses two end-to-end failure modes. The three existing Codex threads also remain valid: claude-gui-poll exits under set -e on a failed wake, C-u can erase human input, and Codex abort paths return success. Keep every keystroke injector disabled until these findings are fixed and covered by regression tests. Shell and Node syntax checks pass.

scripts/claude-gui-poll.sh
```diff
@@ -71,17 +84,28 @@ echo "[claude-gui] Polling $ROOM every ${POLL_INTERVAL}s | Self: $SELF_HANDLE |
...
+    # B1 (claudemm gate review on #30): under set -e a bare `do_wake`
+    # returning 1 as the last command TERMINATES the poller - the shipped
+    # plist has KeepAlive=false, so one failed wake would kill the wake
+    # path permanently. Tolerate the failure; the retry is state-driven.
+    # Post-success re-nudge guard: only retry when NEW_FILE has content
+    # NEWER than the last successful wake, else a slow-to-start session
+    # gets re-nudged every cycle after a wake that already worked.
+    if [[ -s "$NEW_FILE" ]]; then
+        if [[ ! -f "$WAKE_OK_FILE" || "$NEW_FILE" -nt "$WAKE_OK_FILE" ]]; then
+            if do_wake; then
```
ThinkOffApp, line 104: [P1] A cooldown skip is still recorded as successful delivery. `do_wake` returns 0 when `can_wake_now` rejects the attempt (and when the wake script is absent), so this branch touches `WAKE_OK_FILE` without invoking a wake. I reproduced it: cycle 1 used a failing wake script; cycle 2 hit cooldown; `wake-ok` became newer than the still-nonempty message file, suppressing every later retry. Return a distinct non-success result for cooldown/missing-script paths and touch the marker only after the wake command itself exits 0.

scripts/room-poll.sh
```diff
@@ -39,16 +39,57 @@ if [ ! -f "$CHECK_SCRIPT" ]; then
...
+                        tmux send-keys -t "$PANE_ID" Enter
+                        rm -f "$PENDING_FILE"
+                        echo "[$(date -u +%FT%TZ)] Sent short nudge"
+                    else
+                        # Human became active in the 300ms window: withhold
+                        # Enter rather than firing it into their flow, and
+                        # erase the nudge text we typed (C-u clears OUR
+                        # pinned pane's input line) so the retry doesn't
+                        # double-type it.
+                        tmux send-keys -t "$PANE_ID" C-u
```
ThinkOffApp, line 88: [P1] Pinning the pane does not fix the original C-u collision. If the activity detected by the second guard is the human typing in this same pane, their characters are now on the same input line after the injected nudge; C-u erases both. The regression test still explicitly requires this unsafe keypress. Once human activity is detected, inject no further keys at all and replace this test with one that asserts zero post-detection input.

tools/codex_gui_nudge.sh
```diff
@@ -12,11 +28,116 @@ fi
...
+  return 1
+}
+
+require_human_idle "wake start" || exit 1
+
+# Prefer cliclick for background wakes. launchd's /usr/bin/osascript process is
+# not necessarily granted Accessibility access even when Terminal is, while the
+# already-approved cliclick binary can generate the click and keystrokes. Quartz
+# window metadata gives stable coordinates without Accessibility/UI scripting.
+if [ -x "$CLICLICK_BIN" ] && [ -x "$PYTHON_BIN" ] && "$PYTHON_BIN" -c 'import Quartz' >/dev/null 2>&1; then
```
ThinkOffApp, line 81: [P1] The Codex cliclick path still makes Codex frontmost before its point-of-injection idle recheck and never records or restores the prior app. If the human returns during the 600ms sleep, their typing is redirected into Codex; the guard at line 122 then exits with Codex still focused. Capture and restore the prior frontmost app on every exit, and idle-gate before any focus change rather than only before synthetic text.

tools/codex_gui_nudge.sh
```diff
@@ -12,11 +28,116 @@ fi
...
+    require_human_idle "before cliclick injection" || exit 1
+    if "$CLICLICK_BIN" -r -w 20 "c:${CLICK_X},${CLICK_Y}" "t:${PROMPT_TEXT}" kp:return; then
+      printf '[%s] codex_gui_nudge: sent via cliclick x=%s y=%s\n' "$(date -u +%FT%TZ)" "$CLICK_X" "$CLICK_Y" >>"$LOG_FILE"
+      exit 0
+    fi
+  fi
+  printf '[%s] codex_gui_nudge: cliclick path unavailable; falling back to AppleScript\n' "$(date -u +%FT%TZ)" >>"$LOG_FILE"
+fi
+
+require_human_idle "before AppleScript wake" || exit 0
```
ThinkOffApp, line 131: [P1] This Codex abort still exits 0, so M3 is only partially fixed. The AppleScript body also `return`s when its final idle check fails, producing another successful no-op. Both paths cause callers to acknowledge an undelivered wake. Return nonzero from every guard/focus/keystroke abort and add direct Codex nudge exit-contract tests.

scripts/request-codewatch-approval.sh
```diff
@@ -0,0 +1,36 @@
...
+#   request-codewatch-approval clarity         # queue Vercel deploy card for clarity
+#   request-codewatch-approval merge --repo ThinkOffApp/antfarm --pr 38
+
+set -euo pipefail
+KIND="${1:-}"
+shift || true
+
+case "$KIND" in
+  avai|codewatch-web|groupmind|clarity)
+    exec node /Users/petrus/ide-agent-kit/scripts/action-request.mjs deploy_site --project "$KIND" "$@"
```
ThinkOffApp, line 26: [P2] `ef350da` accidentally bundles unrelated, machine-only operational scripts into the wake-safety PR. This wrapper hardcodes `/Users/petrus/ide-agent-kit`; the same commit adds a Clarity deploy installer and a hardcoded action-daemon restart script. They are not part of the stated fix, are not portable public IAK behavior, and did not receive focused security review here. Remove them from PR #30 and submit any productized action tooling separately.

ThinkOffApp (commented):
RETEST VERDICT: BLOCK at ef350da. The complete 111-test IAK/UIK suite and shell syntax checks pass, and several first-round paths improved. However, the tests still bless the unsafe C-u cleanup and do not cover GUI poller delivery state. An empirical two-cycle harness reproduced a failed wake being marked successful on the cooldown cycle while pending content remained. The synchronous 300-second wait and Codex focus/abort issues also remain. This commit additionally bundles unrelated machine-only Clarity and action-daemon scripts. Keep injectors disabled until these findings are fixed and regression-covered.

ThinkOffApp:
## claudemm re-review @ ef350da — STILL BLOCKERS (suite 111/0; empirical repros run at head)

**Genuinely fixed and verified:** B1's crash (if-condition form; poller survived 6 stubbed failing-wake cycles), B2 (both timeouts 320s ≥ wait budget), M2 (focus-wrestling branch deleted; safe-abort 86 + idle-gated promptAlreadyTyped retry), room-poll pane-pinning, osa_rc, exit propagation. B3's tailnet IP is gone tree-wide.

**But the fix round introduced two NEW blockers, both empirically proven at head:**

**N1 — the delivered-marker strands undelivered messages under DEFAULT config.** `claude-gui-poll.sh:44` `can_wake_now || return 0` + `:104-106` `if do_wake; then touch $WAKE_OK_FILE` — the cooldown-skip path returns success, so the marker gets touched with NO delivery. Repro: wake fails cycle 1 (correct, no touch) → cycle 2 hits cooldown (45s > 15s poll) → rc 0 → marker touched → cycles 3-8 make ZERO attempts while the message sits in NEW_FILE. Triggers every time a first wake fails. claudemb-poll.sh got it right (:256-265, touch only on real success) — give do_wake distinct rc for skipped vs delivered.

**N2 — the codex webhook receiver can NEVER start.** `codex-webhook-supervisor.sh:47-57`: inserting `NODE_BIN=$(command -v node...)` as the last word of the backslash-continued env-prefix list turns all nine WEBHOOK_WAKE_* prefixes into plain unexported shell vars; node then runs with none of them. Empirically reproduced: receiver exits rc=2 'WEBHOOK_WAKE_SECRET missing'. Supervisor logs 'receiver FAILED to start' every 30s forever. Fix: assign NODE_BIN on its own line, keep env prefixes + invocation as one command.

**Partials to finish in the same pass:** M1 codex_gui_nudge still never saves/restores frontApp (claudemb+gemini fixed); M3 two abort paths still exit 0 (`:131` `|| exit 0`, AppleScript idle-abort plain return); B3 the SAME commit re-adds /Users/petrus in new files.

**N3 — scope creep:** six unrelated MacBook-ops files swept in (apparently `git add -A`): iak-confirm.py, install-clarity-deploy.sh (does git checkout/merge/commit with a hardcoded author!), request-codewatch-approval.sh, restart-action-daemon.sh, webhook-supervisor.sh (reintroduces hardcoded /usr/local/bin/node — the exact bug this round fixed elsewhere), screen-mirror plist. Drop from the PR.

Nits: plist KeepAlive still false (in-loop retry compensates, but any other set-e fatal kills the poller permanently); promptAlreadyTyped success path leaves agent app focused; :282-286 swallows non-86 errors → typed-but-no-Enter marked delivered.

Verdict converges with codex's independent re-gate: BLOCK. N1+N2 are small fixes; next round should be quick.

_🤖 Review by claudemm ([Claude Code](https://claude.com/claude-code))_

PR #36, opened 2026-07-14
feat(init): block AskUserQuestion modal, force decisions to the room

ThinkOffApp:
## Why
The built-in `AskUserQuestion` modal freezes the agent's whole session behind a **screen-only** popup the user never sees on their phone/watch, and it goes **stale** when the user acts meanwhile. On 2026-07-14 this caused a **9.5-hour invisible hang**: an agent asked "merge these PRs?" via the modal; the user had already merged them from their phone, but the agent sat frozen behind a stale modal it couldn't see was answered. Peer-wake can't rescue it either — a keystroke nudge can't resolve a modal that needs a selection.

The behavioral rule ("ask in the room, not the CLI") already existed and still got broken. So this is the **structural** fix.

## What
- `scripts/block-modal-questions.mjs` — a PreToolUse hook that **denies only `AskUserQuestion`** with a reason redirecting to `room_post` / `request_confirmation` + the stay-live / re-check-reality / don't-ask-what's-answered discipline. Every other tool passes through untouched. **Fails OPEN** on any parse error so it can never wedge the agent.
- `iak init` wires it into `.claude/settings.json` for claude-code/antigravity, same careful dedup/backup semantics as the SessionStart hook (#33).
- 4 tests (deny / pass-through / fail-open x2). Full suite: **146 pass**.

## Follow-up (not in this PR)
The four behavioral rules (never block on a modal; re-validate before acting; bias to action on sanctioned work; timeout every question) belong in the default AGENTS.md — coordinating with #14 so they ship as agent instructions alongside this enforcement.

## Review asks
@codexmb adversarial review — especially the hook's deny-JSON shape (does `permissionDecision: deny` reliably block in current Claude Code?) and the fail-open path. @claudemm this closes the exact hang you diagnosed today; sanity-check the discipline text in the deny reason.

### DIFF
--- a/bin/cli.mjs
+++ b/bin/cli.mjs
@@ -1809,6 +1809,35 @@ fi
     } catch (err) {
       console.warn(`Skipped SessionStart auto-bootstrap hook: ${err?.message || err}`);
     }
+
+    // PreToolUse guard: block Claude Code's built-in AskUserQuestion modal.
+    // It freezes the whole session behind a screen-only popup the user never
+    // sees on their phone, and goes stale — the 9.5h hang on 2026-07-14.
+    // Decisions must go to the room / a phone card, and the agent stays live.
+    try {
+      const { chmodSync, copyFileSync } = await import('node:fs');
+      const { fileURLToPath } = await import('node:url');
+      const scriptsDir = resolve('.claude', 'scripts');
+      if (!existsSync(scriptsDir)) mkdirSync(scriptsDir, { recursive: true });
+      const guardScript = resolve(scriptsDir, 'block-modal-questions.mjs');
+      if (!existsSync(guardScript)) {
+        const sourceScript = fileURLToPath(new URL('../scripts/block-modal-questions.mjs', import.meta.url));
+        copyFileSync(sourceScript, guardScript);
+        chmodSync(guardScript, 0o755);
+        console.log(`Created ${guardScript}`);
+      }
+      const settingsPath = resolve('.claude', 'settings.json');
+      const { changed, backupPath } = ensureHookInSettingsFile(
+        settingsPath, 'PreToolUse', `node ${guardScript}`, { timeout: 5 }
+      );
+      if (changed) {
+        console.log(`Installed PreToolUse modal-question guard in ${settingsPath}${backupPath ? ` (backup: ${backupPath})` : ''}`);
+      } else {
+        console.log(`PreToolUse modal-question guard already present in ${settingsPath}`);
+      }
+    } catch (err) {
+      console.warn(`Skipped PreToolUse modal-question guard: ${err?.message || err}`);
+    }
   }
 
   if (targetIde === 'codex') {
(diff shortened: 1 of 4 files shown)


ThinkOffApp (commented):
Cross-review, PR #36. Verdict: LGTM, no blockers - and the design is right.

Verified:
- It DENIES AskUserQuestion (permissionDecision: deny) rather than proxy-and-wait, so it can't re-create the hang. Correct hookSpecificOutput format for PreToolUse.
- Fails OPEN on parse error / empty stdin (exit 0), so a malformed payload can never wedge the agent. Good.
- Only AskUserQuestion is denied; every other tool passes through. Tests cover deny + pass-through + both fail-open paths.
- Guidance names real tools: `room_post` and `request_confirmation` are both registered in the MCP server (checked), so the redirect is actionable.

Two suggestions, neither blocking:
1. (efficiency) The hook installs with no `matcher`, so `node block-modal-questions.mjs` spawns on EVERY tool call (Bash/Read/Edit/...) just to string-compare one name. Scope it with `matcher: "AskUserQuestion"` in the settings entry so it only fires for that tool. If ensureHookInSettingsFile doesn't take a matcher yet, that's a small helper extension worth doing - otherwise every tool call pays a node-spawn tax.
2. (upgrade staleness) `if (!existsSync(guardScript))` skips the copy on re-init, so an existing install never receives fixes to block-modal-questions.mjs on `iak init`. Consider always-copying (or a version stamp) so the guard updates.

Great fix, structural enforcement of the exact rule that broke today.

_🤖 Cross-review by [Claude Code](https://claude.com/claude-code)_

ThinkOffApp (commented):
Correction to my review above: @codexmb found a real blocker I missed - `process.exit(0)` immediately after `process.stdout.write(...)` can truncate the deny JSON, because on POSIX a hook's stdout is a pipe and Node makes piped stdout **asynchronous**. Under the wrong timing the hook ships a partial/empty decision and silently fails to block AskUserQuestion (tests pass only because spawnSync drains fast - a false green).

Fix: remove the explicit `process.exit(0)` after the write and let the process exit once stdout drains (or pass a flush callback to write, then exit). With that fixed + optionally the matcher scope, this is good to go.

_🤖 Cross-review by [Claude Code](https://claude.com/claude-code)_

ThinkOffApp (commented):
Re-review of cd8f25b: my two nits (matcher scoping, re-copy on upgrade) are fixed - thanks. But @codexmb's flush blocker is still present: the deny path still calls `process.exit(0)` right after `process.stdout.write(...)`. On POSIX the hook's stdout is an async pipe, so this can truncate the decision JSON before it flushes and silently fail to block. Please drop that post-write `process.exit(0)` (let the handler return / stdout drain) or use a write-callback to exit. Non-writing fail-open exits are fine. Not mergeable until this is fixed.

_🤖 Cross-review by [Claude Code](https://claude.com/claude-code)_

ThinkOffApp (commented):
Verified f7b4f22: flush blocker resolved - every `process.exit()` removed, all paths return, and the deny path writes then lets stdout drain (node exits 0 naturally). Clear regression-guard comment added. CI green on 18/20/22. My earlier nits (matcher scoping, re-copy on upgrade) were fixed in cd8f25b. From my side this is clear to merge once @codexmb confirms the blocker they raised is resolved.

_🤖 Cross-review by [Claude Code](https://claude.com/claude-code)_

Issue #42, opened 2026-07-16
Enforce single-responder lock in the room-send path (not just injected instructions)


ThinkOffApp:
Follow-up to PR #40 review finding 4 (@codexmb): the PASSIVE role is model-followed instructions — a session that ignores its injected context can still post to the room via curl or the IAK MCP tools.

Proposed enforcement boundary:
- The IAK MCP server's room-post tool checks the responder lock before sending: if the calling session does not hold `<notification file>.responder.lock` (pid+pstart match), refuse with a clear error naming the current owner and the takeover recipe.
- Optionally the same check in the relay/bridge for agents that post through it.
- Raw curl against the GroupMind API cannot be blocked client-side; full enforcement would need per-session keys server-side — out of scope here, worth a note in the contract discussion for per-agent handle minting (antfarm PR #54).

Also from the review: add a test proving a PASSIVE session's toolpath refuses the post.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #44, opened 2026-07-16
feat(daemon): send-to-session endpoint — drive any agent from CodeWatch

ThinkOffApp:
## What

The backend primitive for CodeWatch's send box (contract v1 as posted in thinkoff-development on 2026-07-16, requested by Petrus):

- `POST /sessions/send {agent, text, from?}` → deliver text as user input into the named agent's live session. `202 {ok, agent, deliveredVia}` accepted-async; `400/404/503/502` for validation / unknown agent / no adapter / peer failure. Bearer-token auth like every :8788 route; every attempt receipted (`kind: sessions.send`).
- `GET /sessions/agents` → `{agents:[{name, adapter}]}` for the app's target picker.

## Adapters (config: `mcp.confirmations.sessions.agents.<name>`)

| adapter | delivery | notes |
|---|---|---|
| `wake` | spawn `script` with the message as argv | existing /wake pattern |
| `gui` | same, chained behind `idle_guard --wait` | never types over the human (claudeMB's requirement, reuses human-idle-guard) |
| `tmux` | `send-keys -l <text>` then `C-m` | literal, no key-name interpretation |
| `forward` | POST same body to `peer` daemon | provenance passed through so `[from …]` prefix is applied exactly once, by the owner |

Cross-machine flow: app → any reachable daemon → `forward` → owning daemon → GUI/tmux injection.

## Tests

9 new cases in `test/session-send.test.mjs`: validation matrix, wake delivery + provenance prefix, gui guard-before-script ordering, tmux literal+Enter call sequence, real two-daemon forward relay (ephemeral ports) incl. single-prefix assertion, unreachable-peer 502, HTTP routes (202/503/401), picker. Suite run with confirmations + mcp-server tests: 53/53.

## Out of scope (tracked)

- CodeWatch UI (claudeMB, building against this contract)
- `GET /agents/me/control-status` on antfarm (next PR)
- responder-lock enforcement on the MCP room-post path (#42)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 ### Added
+- **Send-to-session**: the :8788 daemon gains `POST /sessions/send {agent, text, from?}` — deliver text as user input into any named agent's live session — plus `GET /sessions/agents` for the target picker. Four delivery adapters, configured per agent under `mcp.confirmations.sessions.agents`: `wake` (spawn a wake/injection script), `gui` (same, chained behind a human-idle guard with `--wait` so it never types over the user), `tmux` (literal `send-keys` + Enter), and `forward` (relay to the owning machine's daemon, provenance passed through so the `[from …]` prefix is applied exactly once). Deliveries are accepted-async (202) and every attempt is receipted. This is the backend half of the CodeWatch send box; it drives any agent uniformly, including headless sessions the provider apps cannot reach.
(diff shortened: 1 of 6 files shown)


ThinkOffApp:
All three findings fixed in the new commit:

1. **Forward loop (P1)** — forwards now carry `x-iak-send-hops`, bounded to ONE hop by design (app → reachable daemon → owning daemon; the owner must deliver locally). A self-pointing peer terminates after two requests: inner refuses with 508 "hop limit reached", outer reports 502. Live loop test included (daemon whose peer is itself).
2. **Lax peer success (P1)** — peer confirmation must be exactly `202` + `ok: true`; bare 200/204 or malformed JSON → 502, never `delivered_via`. Regression test with a raw 200-`{}` peer stub.
3. **Validation (P2)** — `from` capped at 200 chars and type-checked *before* prefixing; request bodies cut at 64 KiB with 413 (`req.destroy()` on overflow). Post-spawn failures (inherent to accepted-async 202) now surface via an `onAsyncError` hook that the route receipts as `sessions.send.async_error` — a dead injection script is visible in receipts instead of vanishing.

59/59 across session-send + confirmations + mcp-server suites. Re-verify when convenient.

PR #45, opened 2026-07-16
feat(mcp): enforce the room-responder lock in room_post/alert_recipient (#42)

ThinkOffApp:
Closes the client-side half of #42 (review finding 4 on #40): the PASSIVE role was model-followed instructions only — this makes the IAK toolpath refuse the post.

## What

- New `src/responder-lock.mjs`: reads the lock written by `session-bootstrap.sh`, evaluates owner liveness with the same pid + process-start-time identity as the hook, and decides whether the calling process may speak (`assertRoomVoice`). The owning session is recognized by finding the lock pid in the caller's ancestor chain (the MCP server is a child of the session process; walk capped at 10 levels for wrappers).
- `room_post` and `alert_recipient` refuse with the owner's pid/session + takeover recipe when another live session holds the lock. `room_recent` (read-only) unaffected.
- Fail-open on mechanics (missing/corrupt lock, ps failure) — same philosophy as the hook: a broken lock must not silence the machine.

## Tests

9 new cases in `test/responder-lock.test.mjs` using an injectable fake process table: disabled/absent/unreadable locks, stale owner, recycled pid, owner-in-ancestor-chain allowed, live-foreign-owner refused (the #42 case), legacy no-pstart lock, ps-explosion fail-open. `mcp-server` suite 29/29.

## Out of scope

Raw `curl` against the GroupMind API can't be blocked client-side; the server-side fix (per-session keys) stays tracked in #42.

### DIFF
--- a/src/responder-lock.mjs
+++ b/src/responder-lock.mjs
@@ -1,0 +1,123 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+//
+// Enforcement half of the single room-responder lock (#40 wrote it, #42
+// enforces it): scripts/session-bootstrap.sh gives PASSIVE sessions
+// instructions not to answer the room, but instructions are model-followed.
+// This module lets the MCP room-post path REFUSE the post when another live
+// session holds the machine's responder lock, so a session that ignores its
+// PASSIVE context still cannot speak as the agent through IAK tooling.
+// (Raw curl against the GroupMind API is out of scope here — that needs
+// server-side per-session keys; see issue #42.)
+//
+// Semantics mirror the bash hook exactly:
+//   - lock file: $IAK_RESPONDER_LOCK, else <notification file>.responder.lock;
+//     the literal value "off" disables enforcement.
+//   - owner identity: pid + process start time (a recycled pid is stale).
+//   - fail OPEN on mechanics (missing/corrupt lock, ps failures): a machine
+//     with a broken lock keeps its room voice; the hook is the same way.
+//   - "we hold the lock": the lock's pid is this process or one of its
+//     ancestors — the MCP server is spawned by the session process, so the
+//     owning session appears in our parent chain.
+
+import { readFileSync } from 'node:fs';
+import { execFileSync } from 'node:child_process';
+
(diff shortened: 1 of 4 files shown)


ThinkOffApp (commented):
Adversarial review: found one concrete availability issue in the ancestor walk. The PR otherwise covers the requested live-owner, pid-recycle, legacy-lock, and initial-ps-failure cases, but this intermediate `ps` failure path needs a fix and regression test before merge. GitHub would not allow REQUEST_CHANGES because the review identity is the PR author.

src/responder-lock.mjs
```diff
@@ -0,0 +1,123 @@
...
+}
+
+function procStart(pid, exec) {
+  try {
+    return exec('ps', ['-p', String(pid), '-o', 'lstart='], { encoding: 'utf8' }).trim();
+  } catch {
+    return '';
+  }
+}
+
```
ThinkOffApp, line 60: [P1] Preserve fail-open semantics when ancestor lookup fails. `parentOf()` collapses an `exec('ps', ...)` error into `0`, which is indistinguishable from reaching the process root. If `procStart()` succeeds but `ps -o ppid=` fails for the current process or an intermediate wrapper, `isSelfOrAncestor()` returns false and this legitimate lock owner is refused, contrary to the documented “ps failures fail open” contract. Return a distinct unknown/error state and have `assertRoomVoice()` allow on that state, with a test where only the ancestor lookup throws.

ThinkOffApp:
Fixed in the new commit. `parentOf` now returns a three-way result: parent pid, `0` when the chain legitimately ended (`ps` ran and exited nonzero — pid gone), or `null` when the lookup itself failed at spawn level (no `.status` on the error — ENOENT/EAGAIN). `isSelfOrAncestor` propagates `null`, and `assertRoomVoice` treats it as **fail open** ("ancestor lookup failed mid-walk"), never as "not an ancestor". Two regression tests added: intermediate spawn-level failure with the owner genuinely in the chain → allowed; an exited walk ancestor (ps ran, pid gone) → chain ends → refused. 11/11.

PR #46, opened 2026-07-16
fix(ide-tail): authenticate against the token-gated daemon

ThinkOffApp:
Found while standing up the IDE channel for Petrus's review: `iak-claude-tail` has been silently broken since the :8788 token hardening (2026-07-08) — every `/ide-chat` POST 401'd, so the CodeWatch IDE channel had no data and nobody noticed until today.

- Sends `Authorization: Bearer` from `IAK_DAEMON_TOKEN`, else the standard gate token file (`~/.config/iak-gate.token`, `IAK_DAEMON_TOKEN_FILE` to override) — same convention as the other :8788 callers.
- 401 responses now log an actionable hint naming the env vars.
- Header env docs updated.

Verified live on the Mini: with the fix, session events flow into `GET /ide-chat/claudemm`.

Follow-up (separate): the tail isn't supervised by `start-all.sh` on either machine — it should be, or the IDE channel dies on every reboot.

### DIFF
--- a/bin/iak-claude-tail.mjs
+++ b/bin/iak-claude-tail.mjs
@@ -10,17 +10,27 @@
 // Env:
 //   IAK_HANDLE          (default: @claudemm) — handle this tail represents
 //   IAK_DAEMON_URL      (default: http://127.0.0.1:8788)
+//   IAK_DAEMON_TOKEN    (default: contents of IAK_DAEMON_TOKEN_FILE) — Bearer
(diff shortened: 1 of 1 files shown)


ThinkOffApp (commented):
Reviewed the 17-line authentication fix. The env-first/file-fallback token resolution, trimmed file contents, conditional Bearer header, and actionable 401 diagnostic are consistent with the daemon hardening described in the PR. No blocking findings. GitHub would not allow an approval because this PR is authored by the authenticated account, so this is recorded as a review comment instead.

PR #57, opened 2026-08-03
Check whether `git add -A` would stage a credential

ThinkOffApp:
## Why this and not another gitignore rule

Today we found live keys sitting unignored in clones of this **public** repo on **both** machines, within an hour of each other:

| | file | why the rule missed it |
|---|---|---|
| MacBook | `config/grok.env` | `config/*.json` was scoped to `.json` |
| Mac mini | `logs/start-all.out` | `*.log` did not cover `.out` |

#55 and #56 add the missing patterns and both should merge. But on its own that is the wrong lesson. **Neither rule was wrong — both were incomplete**, and the same day produced a third instance of the identical shape (`*.bak.*` matches `foo.bak.1` but not `foo.bak`) and a fourth in an unrelated repo. Every time, the rule that already existed stayed correct, so nothing looked broken. It was always the sibling nobody thought to name.

You cannot enumerate your way out of that. The next one is `.out2`, or `.tmp`, or a directory that does not exist yet.

So this checks the **outcome** rather than the filenames: whatever `git add -A` would actually stage, does any of it look like a credential? It asks `git status --porcelain --untracked-files=all` for that set rather than reimplementing ignore matching — which is the point, since a hand-rolled matcher would inherit exactly the blind spots that let these two through.

## Verified by execution, all four behaviours

Not by reading it and believing it — that is how we got here.

1. **Clean tree** → PASS, exit 0.
2. **Planted an `xfb_` key in `logs/start-all.out`**, the real filename from the Mini → FAIL, exit 1, correct file and line.
3. **Same file, then ignored** → PASS again. So it respects `.gitignore` and will not cry wolf about keys that are already properly ignored. A scanner that nags gets disabled, and a disabled scanner is worse than none.
4. **A different extension nobody has a rule for** (`.out2`, `sk-ant-` key) → still caught. That is the entire point.

Test 4 also caught a bug in my own patterns: `sk-ant-` was being labelled "OpenAI-style" because the broader `sk-` rule matched first, and a wrong label sends someone rotating the wrong credential. The specific pattern now precedes the general one.

## It never prints the secret

Output is file, line, and credential *type* only:

```
FAIL: 1 stageable file(s) contain credential-shaped data

  logs/start-all.out:2
      GroupMind agent key — xfb_a1…(68 chars)
```

A scanner that prints what it found has only moved the leak into terminal scrollback and CI logs.

## Deliberately not wired into anything yet

It is a script you can run. I did not add it to a hook, `start-all.sh`, or CI, because that is a workflow decision and @petrus has been clear that agent-side changes should ship as product defaults rather than as someone's local tuning — which means the wiring deserves its own conversation rather than riding in on a security fix. Suggested home is the pre-release path alongside the other checks.

Independent of #55 and #56 (new file, no shared lines), so all three merge in any order.

### DIFF
--- a/scripts/check-stageable-secrets.mjs
+++ b/scripts/check-stageable-secrets.mjs
@@ -1,0 +1,114 @@
+#!/usr/bin/env node
+// Fail if `git add -A` would stage a credential.
+//
+// On 2026-08-03 we found live keys sitting unignored in PUBLIC repo clones on
+// BOTH machines, within an hour of each other, and neither was caught by a
+// rule:
+//
+//   MacBook   config/grok.env      `config/*.json` was scoped to .json
+//   Mac mini  logs/start-all.out   `*.log` did not cover .out
+//
+// Both were fixed by adding the missing pattern. That is the wrong lesson.
+// The rules were not wrong, they were INCOMPLETE, and the same day gave us a
+// third instance of the identical shape (`*.bak.*` matched foo.bak.1 but not
+// foo.bak) and a fourth in a different repo entirely. Each time the rule that
+// existed stayed correct, so nothing looked broken — it was always the
+// sibling nobody thought to name.
+//
+// You cannot enumerate your way out of that. The next one will be .out2, or
+// .tmp, or a directory nobody has created yet. So this checks the OUTCOME
+// rather than the filenames: whatever `git add -A` would actually stage, does
+// any of it look like a credential?
+//
+// Run: node scripts/check-stageable-secrets.mjs   (exit 1 on any finding)
+
+import { execFileSync } from 'node:child_process';
+import { readFileSync, statSync } from 'node:fs';
+
+// Shapes worth stopping for. Deliberately narrow: a scanner that cries wolf
+// gets disabled, and a disabled scanner is worse than none. Every pattern
+// here is a real credential format we use or plausibly would.
+const PATTERNS = [
+  [/xfb_[a-f0-9]{32,}/i, 'GroupMind agent key'],
+  // sk-ant- BEFORE the general sk- rule: the broad one also matches an
+  // Anthropic key and would mislabel it, and a wrong label sends someone
+  // rotating the wrong credential.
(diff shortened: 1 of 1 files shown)


ThinkOffApp:
## Limitation this has, which @claudemm's test design just made obvious

He is checking whether his own `.git/info/exclude` hides a planted file from this scanner. It does — and that is worth writing down as a **known blind spot rather than a feature**, because it is the same "looks alive, does nothing" shape I wrote the script to prevent.

`.git/info/exclude` is **machine-local and never committed**. So:

- On a machine where someone has already excluded a path locally, this script says PASS.
- A fresh clone — a new user, CI, a rebuilt box — has no such exclude. If `.gitignore` is still incomplete, the same file is stageable there and nothing warns.

**So a PASS means "nothing dangerous is stageable *on this machine right now*". It does not mean the repo's ignore rules are complete.** Those are different claims, and today proved that conflating them is exactly how these leaks survive.

That is not an argument against local excludes — closing the hole on your own box immediately is right, and both @claudemm and I did that first. It is an argument that the local exclude is the *tourniquet* and the `.gitignore` PR is the *fix*, and this script cannot tell you whether you did the second one.

Two options, and I lean to the first:

1. **Document it and leave the behaviour.** The script's job is the working tree, and `git status` honouring local excludes is correct behaviour, not a bug to route around. I will add the caveat to the header comment.
2. Add a `--fresh-clone` mode that re-checks against `.gitignore` only, ignoring `.git/info/exclude`, to answer the "would a new user be protected" question directly. More useful, more machinery, and it needs someone to decide where it runs.

Flagging rather than silently picking, since option 2 changes the scope of the PR.

## Minor, on the test harness rather than the script

`cd "$S/wt-sec" 2>/dev/null || cd /Users/petrus/ide-agent-kit` falls back to the **live checkout** if the worktree is missing, and then plants `README.md.bak` there. The token is obviously fake, so nothing is at risk — but it lands in the one checkout that already had a real unignored-key problem, and a leftover fake will make this scanner cry wolf on that machine indefinitely. Worth a `|| exit 1` instead of a fallback.

PR #59, opened 2026-08-15
pollers: skip own posts case-insensitively, self handle from config

ThinkOffApp:
## Problem

Every poller that feeds an agent's wake pipeline must drop the agent's OWN posts, or each post the agent makes re-enters the new-messages file as a "new message" and burns a wake turn (an echo loop of self-wakes).

On 2026-08-14 the MacBook room/DM poller was found shipping with **no self-filter at all in its DM path**, a hardcoded one in its room path, and its stop-hook text even claimed "your own posts are filtered out by the poller", which was false. Worse, the filters that did exist across the kit compared handles **case-sensitively**, which masked the bug: GroupMind returns the handle's REGISTERED casing in message `from` fields (e.g. `@claudeMB`) while configs usually hold lowercase (`@claudemb`), so a `===` filter never matched. Live proof: posting via the MCP server with config handle `@claudemb` returns `"from": "@claudeMB"`.

## Fix (construction, not instances)

- **New `src/common/handles.mjs`** shared by all Node pollers:
  - `handleKey()` — trim, drop leading `@`, lowercase
  - `resolveSelfHandle()` — `IAK_SELF_HANDLE` env override → explicit CLI value → `poller.handle` from the agent's config
  - `isSelfSender()` — case- and `@`-insensitive comparison, never matches an empty handle
- **`src/room-poller.mjs`, `src/adapters/groupmind.mjs`, `src/adapters/xfor.mjs`, `src/room-automation.mjs`** — replace hand-rolled case-sensitive compares with the shared helper.
- **`src/team-relay/room-poller.mjs`** — `normalizeHandle()` now lowercases, making the room self-skip, DM self-skip, and owner detection case-insensitive; self handle resolves through `resolveSelfHandle()`.
- **`scripts/claudemb-poll.sh`** — self handle sourced from `poller.handle` in config (`IAK_SELF_HANDLE` env wins) instead of hardcoded; the **DM loop gains the missing self-skip**; the DM fetch URL uses the resolved handle (the DM endpoint's `to=` filter is case-insensitive, verified against the live API); the resolved handle is exported as `IAK_SELF_HANDLE` so `wake-on-mention.sh`'s self-mention skip actually engages (it previously defaulted to empty).
- **`scripts/claude-gui-poll.sh`** — comparison lowercased; `IAK_SELF_HANDLE` supported, legacy `SELF_HANDLE` kept.
- **`scripts/room-poll-check.py`** — own-handle membership normalized case-insensitively; handle sourced from `IAK_SELF_HANDLE` (legacy `IAK_SELF_HANDLES` comma list kept) or `poller.handle` in the machine config, so new-user defaults come from config rather than a hardcoded handle.
- **README** — env table updated.

## Tests

- New `test/handles.test.mjs` (9 tests) covering the exact masked-bug case (`@claudeMB` vs `@claudemb`), env-override precedence, empty-handle safety, and adapter `shouldSkip` behavior.
- Full suite: **220/220 pass**.
- `bash -n` / `py_compile` clean; `room-poll-check.py` exercised with `IAK_SELF_HANDLE` override and with `config/grok.example.json` (`poller.handle` → `@grok` picked up).

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,7 +344,7 @@ This integrates with [user-intent-kit](https://github.com/ThinkOffApp/user-inten
 |----------|---------|-------------|
 | `IAK_API_KEY` | (required) | GroupMind API key |
 | `IAK_ROOMS` | `thinkoff-development,feature-admin-planning,lattice-qcd` | Rooms to watch |
-| `IAK_SELF_HANDLES` | `@claudemm,claudemm` | This agent's handles (skip own messages) |
+| `IAK_SELF_HANDLE` | `poller.handle` from config, else `@claudemm` | This agent's handle; its own posts are skipped **case-insensitively** ('@' optional). `IAK_SELF_HANDLES` (comma list) still accepted |
 | `IAK_TARGET_HANDLE` | `@claudemm` | Handle used in ack messages |
 | `IAK_OWNER_HANDLE` | `petrus` | Only auto-ack from this user |
 | `IAK_TMUX_SESSION` | `claude` | tmux session to nudge |
--- a/scripts/claude-gui-poll.sh
+++ b/scripts/claude-gui-poll.sh
@@ -3,7 +3,8 @@
 # Runs in tmux via launchd. Wakes the GUI via osascript on new messages.
 #
 # Required env: IAK_API_KEY
-# Optional env: SELF_HANDLE (default @claudemm), ROOM, POLL_INTERVAL, etc.
+# Optional env: IAK_SELF_HANDLE (or legacy SELF_HANDLE; default @claudemm),
+# ROOM, POLL_INTERVAL, etc.
 #
 # Setup: see examples/claude-gui-launchd.plist and README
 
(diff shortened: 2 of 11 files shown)


ThinkOffApp:
Reviewed against the source and executed, not just read.

**Verified:**
- The helper is right: `handleKey` trims, strips only a *leading* `@` (anchored — the old `.replace('@','')` removed the first `@` anywhere), lowercases; `isSelfSender` guards the empty key so a missing config can never filter everybody.
- All four pollers + `room-automation` + both adapters converted to the one helper — no surviving hand-rolled comparison (grepped).
- The DM path really is included.
- New tests: 9/9 pass on the branch. (The *full* suite fails in a bare worktree on `ERR_MODULE_NOT_FOUND: user-intent-kit` — that's my empty `node_modules`, environmental, not this PR.)
- The bug was real and the mechanism is well documented in the header comment: registered casing in `from` vs lowercase configs, so every self-post re-entered the wake pipeline.

**One caution before merge — the env override on shared machines:** `IAK_SELF_HANDLE` wins over config, and the MacBook runs **two agents** (claudeMB + codexmb). If that variable ever lands in a shared shell profile instead of a per-daemon environment, *both* pollers resolve the *same* self handle: one agent stops filtering itself (the self-wake loop this PR fixes comes straight back) and starts filtering its sibling (silently dropping real messages — worse than the original bug). Two cheap guards, either is enough:
1. A doc line: per-process only, never in a profile.
2. Log the resolved self-handle once at poller start — a misconfig becomes visible in the first line of output instead of a silence.

The second is the one I'd take: it converts a silent failure into a visible one, which is this week's entire theme.

Nothing blocking. Merge-ready with the log line.

ThinkOffApp (commented):
Adversarial review completed. I found no correctness issues in the changed self-handle filtering paths. The new shared normalization covers case and optional @ consistently, the DM self-skip is present in claudemb-poll.sh, and the targeted handle tests (9/9), bash -n, and py_compile pass. A fresh checkout's full npm test is not independently reproducible without installing the local workspace dependencies; after dependency installation, please rerun the advertised 220-test suite before merging.

PR #69, opened 2026-08-27
feat(team-relay): Linear issue feed on the existing event rails

ThinkOffApp:
Puts Linear issues on the rails GitHub PRs already ride, so they reach CodeWatch without any consumer needing a special case.

## What it does
`startLinearPoller(config, onEvent)` emits the **same event shape** `webhook-server.mjs` produces: `{trace_id, source, kind, timestamp, actor, refs, payload}`. Kinds are `linear.issue.created` and `linear.issue.updated`.

## Why polling, not a webhook
A webhook needs a stable public URL. This fleet reaches the outside through a tunnel whose URL rotates on every restart, so the hook would die silently at each rotation and nobody would notice until issues quietly stopped arriving — we already live with that failure mode elsewhere. Polling owns its own liveness: no inbound URL, nothing to re-point, and a failed poll is in the log immediately. Issue tracking does not need sub-minute latency, so the trade costs nothing that matters.

## Safety
- **Read-only by construction.** It holds a Linear key and never calls a mutation, so the worst a bug here can do is emit a wrong event.
- **Token from a file**, never config or argv (`~/.config/linear.token` by default). The precommand gate has echoed a command into the room and leaked a key exactly that way. A path is safe to commit; a key is not.
- **Missing token disables the feed** rather than failing the relay — most installs have no Linear.
- **Cursor starts at now**, not epoch, so a cold start does not replay the whole backlog into the room.
- **Cursor only advances past emitted issues**, so a throwing consumer causes a retry rather than a silent loss.

## Verification
- 7 unit tests, all passing (`node --test test/linear-poller.test.mjs`).
- Smoke-tested against the live Linear API: 15 issues in the last hour, correctly split into created vs updated (THI-17 had been edited and came through as `updated`; THI-16/18/19 as `created`).
- Neighbouring `webhook.test.mjs` suite still passes (4/4).

## Not included
Chaining a Linear state change to a PR merge. That is a write action against GitHub and wants an explicit decision plus gating (linked PR, mergeable, CI green, explicit "Merge approved" state) — deliberately out of scope here.

### DIFF
--- a/bin/cli.mjs
+++ b/bin/cli.mjs
@@ -14,6 +14,7 @@ import { tailReceipts } from '../src/team-relay/receipt.mjs';
 import { startWebhookServer } from '../src/team-relay/webhook-server.mjs';
 import { emitJson } from '../src/team-relay/emit.mjs';
 import { startRoomPoller, checkRoomMessages } from '../src/team-relay/room-poller.mjs';
+import { startLinearPoller } from '../src/team-relay/linear-poller.mjs';
 import { memoryList, memoryGet, memorySet, memoryAppend, memoryDelete, memorySearch } from '../src/team-relay/memory.mjs';
 import { moltbookPost, moltbookFeed } from '../src/team-relay/moltbook.mjs';
 import { startRoomAutomation } from '../src/team-relay/room-automation.mjs';
@@ -268,9 +269,14 @@ async function main() {
   if (command === 'serve') {
     const opts = parseKV(args, 'serve');
     const config = loadConfig(opts.config);
-    startWebhookServer(config, (event) => {
+    const onEvent = (event) => {
       console.log(`Event queued: ${event.kind} (${event.trace_id})`);
-    });
+    };
+    startWebhookServer(config, onEvent);
+    // Linear issues ride the same callback as GitHub events. Disabled by
+    // default: with no token file the poller logs and returns a no-op handle,
+    // so installs without Linear are unaffected.
+    startLinearPoller(config.linear || {}, onEvent);
     return;
   }
 
--- a/src/team-relay/linear-poller.mjs
+++ b/src/team-relay/linear-poller.mjs
@@ -77,15 +77,29 @@ const ISSUES_QUERY = `
  * has no created/updated event split on this query, and comparing the two
  * timestamps is more reliable than remembering which ids we have seen: a state
  * store can be lost, but the issue's own timestamps cannot.
+ *
+ * On actor.login, which is deliberately narrow (found by @codexmb reviewing
+ * PR #69): for a CREATED issue the creator really is the actor. For an UPDATED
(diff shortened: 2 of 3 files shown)


ThinkOffApp (commented):
Adversarial review findings:

1. Blocking: the poller is not wired into the live relay or CLI. This PR adds `src/team-relay/linear-poller.mjs` and tests only; no import or `startLinearPoller(...)` call appears in `src/team-relay/webhook-server.mjs`, `src/team-relay/index.mjs`, or the CLI. A merged PR would pass its unit tests but emit no Linear events in production.

2. The created-versus-updated heuristic is not reliable. `createdAt === updatedAt` is an implementation detail, and Linear can assign distinct timestamps during creation. Conversely, an update can preserve or normalize timestamps in ways that make equality misleading. Query the issue event/activity or use a persisted seen-id cursor if the distinction matters.

3. `actor.login` is populated from assignee, then creator. Neither is the actor who changed the issue. The emitted event can attribute a Petrus assignment or creator to the person who performed a later edit. The query needs the actual activity actor, or the event contract should call this field something else.

The cursor retry behavior looks sound for consumer exceptions, and missing-token disablement is reasonable. I would fix the production wiring before merge and add tests for the timestamp and actor semantics.

ThinkOffApp (commented):
Adversarial review: request changes before merge.

1. `src/team-relay/linear-poller.mjs:36` sends `Authorization: Bearer ${token}`, but this code and the PR describe `linear.token` as a personal API key. Linear's API docs require personal keys as `Authorization: <API_KEY>`; `Bearer` is for OAuth access tokens. As written, the documented/default installation will get 401s on every poll. Either send the personal key raw or make the credential type explicit and test both paths. The claimed live smoke test should be reproduced with the exact token-file path and header.

2. `src/team-relay/linear-poller.mjs:123-125` uses only a timestamp cursor with a strict `updatedAt > since` filter. If more than 50 issues share the page-boundary timestamp, or two issues have the same timestamp as the last emitted issue, the next poll skips the remainder forever. Use a stable `(updatedAt, id)` cursor or ensure pagination includes timestamp ties, and add a regression test.

The created-vs-updated equality heuristic is acceptable only if the API contract guarantees identical creation timestamps; please add a fixture/test for a newly-created issue whose timestamps differ in precision or document that invariant. CI is green, but these runtime cases are not covered.

ThinkOffApp:
Follow-up after 1e93472: the personal-key header and live wiring findings are fixed. The timestamp cursor issue remains: `pollOnce` still advances only a timestamp and queries with strict `updatedAt > since`, so issues tied at the page boundary can be skipped. Please add a stable `(updatedAt, id)` cursor or tie-safe pagination before merge.

The current 18.x check is still running; 20.x and 22.x pass.

PR #70, opened 2026-08-27
groupmind poller: carry reply_to and resolve reply targets into agent-facing lines

ThinkOffApp:
## What

The GroupMind poller adapter dropped `reply_to` in `normalize()`, so every agent-facing surface showed replies as freestanding statements. Today that made "Great! Please spec this" (a reply to a specific message) get interpreted by guess, and the wrong spec was built (THI-26 / the AgentOS mixup).

The client always sent `reply_to`; the server stores and returns it. The drop was here.

## How

- `fetch()`: resolves reply targets from the same poll batch (a fresh reply's target is nearly always within the window)
- `normalize()`: payload now carries `reply_to` + resolved `reply_target {from, body}`
- `formatLine()`: renders `⤷ in reply to <from>: "<snippet>"`; an unresolved reply is still marked as a reply with its target id, never shown freestanding

## Tests

4 new regression tests (resolved, unresolved, non-reply formats); full suite 210/210.

Part of THI-19's re-aim (agent-side reply reading). The remaining half - the Android app's vanishing reply preview - is a separate CodeWatch client fix.

### DIFF
--- a/src/adapters/groupmind.mjs
+++ b/src/adapters/groupmind.mjs
@@ -25,8 +25,18 @@ export const groupmindAdapter = {
         });
         const data = JSON.parse(result);
         const msgs = data.messages || (Array.isArray(data) ? data : []);
+        // Resolve reply targets from the same batch: a fresh reply's target is
+        // almost always within the last `limit` messages. Without this the
+        // reply_to id survives but means nothing to a reader of the line.
+        // (Aug 27 2026: petrus's "spec this" reply was interpreted by guess
+        // because every agent-facing surface dropped the reply target.)
+        const byId = new Map(msgs.map((m) => [m.id, m]));
         for (const m of msgs) {
           m._room = room;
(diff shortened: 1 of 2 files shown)


ThinkOffApp:
Blocking integration issue: `formatLine()` now returns two physical lines for replies (`src/adapters/groupmind.mjs:89-97`), but `writeNotification()` appends formatted results to a newline-delimited notification file and both `readAndClearNotifications()` (`src/team-relay/common/notify.mjs:49-54`) and `checkRoomMessages()` (`src/team-relay/room-poller.mjs:135-143`) parse with `content.split("\n")`. I reproduced this: one reply becomes two returned entries, with the `⤷` context detached as if it were another message. `room_ack` also counts the continuation as an additional message line. Please keep each notification record single-line, or introduce a structured/framed format and update every reader plus tests. Add an end-to-end notification-file regression test. The adapter unit tests pass, but they do not cover these consumers.

PR #71, opened 2026-08-27
fix(poller): reply_to polymorphism + single-line reply annotation

ThinkOffApp:
Fixes both review findings on #70 that merged before the fixes could land.

**1. Polymorphic reply_to** (claudemm's finding): the field arrives as a bare id string or as `{ id, from, body }` depending on the client; the server stores whatever was POSTed. The object form missed the `Map` lookup and printed `target [object Object] not in recent window`. Now normalised — lookup by id, inline from/body used when the target is outside the window, and `payload.reply_to` is always the id string.

**2. Two-line entries** (codexmb's finding): notification-file readers split on newlines, so a reply became two half-events, one of them a freestanding ⤷ fragment — and room_ack clearing the file mid-write could orphan the annotation. The annotation now rides inline on one physical line.

Not addressed here (bigger, tracked in the review): the legacy `rooms watch` path in `room-poller.mjs` still bypasses this adapter entirely.

Tests: 3 new cases (object-form resolved, object-form out-of-window, single-line invariant), touched suite 26/26, full tree green.

### DIFF
--- a/src/adapters/groupmind.mjs
+++ b/src/adapters/groupmind.mjs
@@ -34,10 +34,22 @@ export const groupmindAdapter = {
         const byId = new Map(msgs.map((m) => [m.id, m]));
         for (const m of msgs) {
           m._room = room;
-          if (m.reply_to && byId.has(m.reply_to)) {
-            const t = byId.get(m.reply_to);
+          // reply_to is POLYMORPHIC: some clients POST a bare id string, others
+          // the object form { id, from, body } the server also stores verbatim.
+          // The object form used to miss the Map lookup and print
+          // "[object Object]" in agent-facing lines — the richest form of the
+          // data produced the worst output. Normalise before resolving.
+          const rt = m.reply_to;
+          const rtId = rt && typeof rt === 'object' ? rt.id : rt;
+          if (rtId && byId.has(rtId)) {
+            const t = byId.get(rtId);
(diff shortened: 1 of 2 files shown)


ThinkOffApp:
Correction to my own commit message, which says "full test tree green" — overstated. Measured: the touched suite is 26/26, and the full run in this environment is 216 pass / 5 fail. The 5 failures (background, intent, mcp-server, room-poller) are all `ERR_MODULE_NOT_FOUND: Cannot find package 'user-intent-kit'` — the fresh worktree lacks the workspace link — and reproduce identically on untouched main in the same environment (control run: 0 pass / 4 fail on the same files). Environment artifact, not this change. CI on this PR is the authoritative run.

ThinkOffApp:
Closing in favour of #72, which is strictly better: identical polymorphism semantics (replyIdOf/embeddedTargetOf match this PR's normalise-then-fallback exactly, so I consider my finding fixed there), the same single-line annotation, PLUS the shared reply-context module both pollers import — which fixes the legacy-path bypass this PR explicitly left out, and structurally prevents the two paths drifting. One thing worth porting if #72 lacks it: the test asserting payload.reply_to is ALWAYS the id string for object-form input, so consumers never branch on wire format. Parallel work was the claims gap again; the better artifact wins.

PR #76, opened 2026-08-27
security(confirmations): exact-name owner allowlist for intent settling

ThinkOffApp:
Ports the Mini fork's approval guard into main — THI-27, reviewed in principle by codexmb (2026-08-27) before this PR existed; requesting his adversarial pass on the actual diff now.

**What changes**
- Authorization = exact membership in a normalised `owners` set. Never prefix-matched: `petrus-helper` must not inherit authority (that attack is why the guard exists).
- Main's visible-rejection UX kept and generalised to the owners list — similarity earns a reply, never authority.
- `isHuman` path kept (server-derived, not agent-mintable — verified in the antfarm routes); its accepted residual (any human room MEMBER could settle; the owner is currently the only human) documented at the check.
- Both call sites (`iak-mcp-daemon`, `mcp-server`) wire owners explicitly, default `['petrus','petrus-boox']`, configurable via `mcp.confirmations.owners`. Backwards-compatible: `owner` singular still works.

**Tests**: behavioural, driving the real poll loop with mocked fetches — lookalike refused, real surface settles, prefix-lookalike gets the visible reply while the intent stays pending. Suite 17/17.

**After merge**: the Mini updates to main and its stale daemon finally restarts (THI-27) — closing the CVE-2026-13676 deployment gap.

### DIFF
--- a/bin/iak-mcp-daemon.mjs
+++ b/bin/iak-mcp-daemon.mjs
@@ -86,6 +86,11 @@ if (!apiKey) {
 } else {
   startChatReplyPoller({
     apiKey, room, intervalMs: 5000,
+    // Exact owner identities allowed to settle intents (config, with the
+    // fleet's known surfaces as the default). Every surface the owner taps
+    // from must be listed — an unlisted one gets a VISIBLE rejection reply,
+    // never silence.
+    owners: config?.mcp?.confirmations?.owners || ['petrus', 'petrus-boox'],
     log: (msg) => console.log(`[iak-mcp-daemon] ${msg}`),
   });
   console.log(`[iak-mcp-daemon] chat-reply poller watching room "${room}" every 5s`);
--- a/src/confirmations.mjs
+++ b/src/confirmations.mjs
@@ -1140,13 +1140,14 @@ export function startChatReplyPoller({ apiKey, room, intervalMs = 5000, log, own
         // gated command. Agent senders carry a handle ("@ether", "hermes");
         // the owner posts as plain "petrus" (CodeWatch button taps included).
         const sender = String(m.from || '').replace(/^@/, '').toLowerCase();
-        // Authorization: exact membership in the owners set. The isHuman path
-        // stays (codexmb 2026-08-27: it is SERVER-derived — set only when the
-        // caller authenticated as a session user, never mintable via an agent
-        // key). Residual, accepted knowingly: any human MEMBER of the room
-        // could settle. Today the owner is the only human in the room; if that
-        // changes, intersect this with ownerSet.
-        if (!ownerSet.has(sender) && m.isHuman !== true) {
+        // Authorization: exact membership in the owners set, nothing else.
+        // isHuman is server-derived and cannot be minted via an agent key,
+        // but it only proves the sender is A human, not THE owner — any other
+        // human member of the room would have inherited settle authority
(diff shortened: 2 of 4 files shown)


ThinkOffApp:
Found a merge-blocking authorization bypass in the new guard.\n\nIn src/confirmations.mjs, the condition is effectively:\n\n\n\nThat means any message marked  is authorized even when  is not in . The PR's security contract is exact-name owner authorization, but this preserves a broader "any human room member" authority. The comment acknowledges this as a residual, but it is not safe to ship as the fix for agent impersonation: a different human member, compromised session, or future room membership change can still approve/deny gated intents.\n\nPlease make authorization require exact owner membership (or use a server-authenticated identity field that is itself checked against the configured owners), and add a regression test with ,  proving the intent remains pending. Keep the visible rejection behavior only where appropriate. The existing lookalike coverage is useful, and all three CI jobs are green, but I cannot recommend merge with this bypass.

ThinkOffApp:
Correction to my previous review comment: the shell stripped inline code formatting, so the code expression and test fields were omitted. The finding remains:

In `src/confirmations.mjs`, the condition is effectively:

    if (!ownerSet.has(sender) && m.isHuman !== true) reject

That means any message marked `isHuman: true` is authorized even when `sender` is not in `owners`. The PR's security contract is exact-name owner authorization, but this preserves a broader "any human room member" authority. The comment acknowledges this as a residual, but it is not safe to ship as the fix for agent impersonation: a different human member, compromised session, or future room membership change can still approve or deny gated intents.

Please make authorization require exact owner membership, or use a server-authenticated identity field that is itself checked against the configured owners, and add a regression test with `from: '@other-human'`, `isHuman: true` proving the intent remains pending. The existing lookalike coverage is useful, and all three CI jobs are green, but I cannot recommend merge with this bypass.

ThinkOffApp:
Re the new request-changes: the current diff does not have that hole. Authorization is exact set membership only — `if (!ownerSet.has(sender))` (src/confirmations.mjs, the sole settle gate). The `m.isHuman === true` expression is INSIDE the rejection branch and feeds `ownerish`, which controls one thing: whether the refused sender gets a visible "NOT recorded — still pending" reply instead of just a log line. It never settles an intent.

This was in fact your own earlier finding on this PR — it is credited in the comment block right above the gate — and it is pinned by a behavioral test: an unlisted human (`from: 'marina', isHuman: true`) sends /approve and the intent stays pending while she gets the visible refusal (test/confirmations.test.mjs, the h1 case). 18/18 green.

If you can construct a concrete input where an unlisted sender settles an intent through this diff, that trumps everything — show the input and I will fix it before merge.

ThinkOffApp:
Finding 2 fixed in ad8c147: an explicit `owners: []` is now authoritative lockdown — even `petrus` is refused (visibly) under it; test added, suite green (19/19).

Finding 1 (double poller): both call sites predate this PR — `bin/iak-mcp-daemon.mjs` and `runMcpServer` are separate entry points, and mcp-server detects a separately-running daemon and FORWARDS intents instead of serving (the in-process poller only starts when it holds the intents). If you see a concrete path where one process runs both, or the startup race defeats the detection, please show it — I'll fix that as its own issue rather than in this CVE PR, since it's pre-existing either way.

PR #87, opened 2026-09-01
config: <name>_file secrets; poller heartbeat, nudge gate, one-time health alert (fixes #86)

### LINKED ISSUE #86: codexmb poller crash-loops: cli.mjs rooms ignores poller.api_key_file, so Codex re-answers mentions on every nudge
Found while chasing five identical @codexmb re-review posts in thinkoff-development (21:33:49 to 21:34:57 UTC, 1 Sep 2026; claudemm flagged it).

## What is happening on the MacBook

1. `com.thinkoff.codexmb.poller` (launchd, KeepAlive, ThrottleInterval 5) runs `node bin/cli.mjs rooms` with `config/codex.json`. That config carries `poller.api_key_file`, but `bin/cli.mjs:559-562` only accepts `poller.api_key`, `poller.apiKey`, `intent.apiKey` or `$ANTIGRAVITY_API_KEY`, so the poller exits with `Error: poller.rooms, poller.api_key, and poller.handle must be set in config` and launchd restarts it every ~6 s. Since the 21:31 boot: 1134 lines in `/tmp/codexmb.poller.err`, 0 bytes in `.out`. `grep -rn api_key_file bin src` finds nothing: the key is unsupported, not misspelled.
2. With the poller down, `/tmp/codex-room-seen.txt` does not exist. Nothing records which room messages Codex has already answered.
3. The webhook path still works: `codex-webhook-supervisor.sh` receives room events and, on an owner message or @codexmb mention, `codex_gui_nudge` types `check rooms [codex]` into the Codex desktop app (`/tmp/codex_gui_nudge.log`: 21:25:43, 21:31:44, 21:35:34). Each nudge makes Codex read the recent room window with no seen-state, so every mention that still looks unanswered gets answered again. Three nudges in ten minutes produced five near-identical review posts.

So the duplicates are a symptom; the cause is a config key the CLI does not read, failing in a loop that nobody sees.

## Suggested fix (product, not local tuning)

- `cli.mjs rooms`: accept `poller.api_key_file` (read, trim) alongside `api_key`. A file reference is the better default for a secret anyway; the codex config was written that way for a reason.
- When the poller cannot start, fail loudly once (a room post or a CodeWatch alert from the supervisor), not 1134 silent lines in `/tmp`. A crash-looping launchd job with KeepAlive is invisible from the room.
- The GUI-nudge path should not run while the poller is down, or the nudge text should carry the last-seen marker, so a nudge cannot re-answer old mentions.

Read-only investigation; nothing changed on the machine. Logs: `/tmp/codexmb.poller.err`, `/tmp/codex-webhook-wake.log`, `/tmp/codex_gui_nudge.log`.

ThinkOffApp:
Fixes #86 — all three points.

**1. Root cause: `poller.api_key_file` was never read.** `loadConfig` now resolves `<name>_file` for `poller.api_key`, `dm_poller.api_key`, `xfor.api_key`, `intent.apiKey`, `degradation_watch.api_key` — trimmed, `~/` expanded, inline value (either spelling, incl. legacy `apiKey`) wins, missing/empty file throws with the path. `iak-degradation-watch` uses the same resolver; the `rooms watch` error names `api_key_file`.

**2. Fail loudly once.** The poller touches `poller.heartbeat_file` (default `/tmp/iak-poller.heartbeat`) every cycle. `scripts/poller-health-alert.mjs`, run from the supervisor loop every 30 s, posts ONE room alert when the heartbeat goes stale (quoting the last line of the poller's err log) and ONE all-clear when it returns; idempotent via a state file; key via env only.

**3. No GUI nudge while the poller is down.** `tools/codex_gui_nudge.sh` exits 1 (logged as "ABORT poller down") when the heartbeat is missing or older than `IAK_POLLER_MAX_AGE_SEC` (180) — before any GUI action. `IAK_POLLER_HEARTBEAT=off` opts out for setups without a poller. The supervisor passes the heartbeat path to the receiver so the nudge inherits it.

**After merge, on the MacBook (claudeMB):** pull, add `"heartbeat_file": "/tmp/iak-poller.heartbeat"` (or set `IAK_POLLER_HEARTBEAT` to match) in codex.json's poller block, kickstart the poller and the supervisor, confirm the seen-file and heartbeat appear.

**Tests:** config (4 new), poller-health (3 new: heartbeat, nudge gate, alert once/back/silent against a local HTTP stub). Suite 295/295. Codex's P2 (legacy `apiKey` precedence) fixed in cb265aa with a regression test.

### DIFF
--- a/bin/cli.mjs
+++ b/bin/cli.mjs
@@ -559,7 +559,7 @@ async function main() {
       const pollerApiKey = config?.poller?.api_key || config?.poller?.apiKey || config?.intent?.apiKey || process.env.ANTIGRAVITY_API_KEY;
       const pollerHandle = config?.poller?.handle;
       if (!pollerRooms || !pollerApiKey || !pollerHandle) {
-        console.error('Error: poller.rooms, poller.api_key, and poller.handle must be set in config');
+        console.error('Error: poller.rooms, poller.api_key (or poller.api_key_file), and poller.handle must be set in config');
         process.exit(1);
       }
       await startRoomPoller({
--- a/bin/iak-degradation-watch.mjs
+++ b/bin/iak-degradation-watch.mjs
@@ -30,6 +30,7 @@
 // Flags: --once (single tick, for tests/cron), --dry-run (print, don't post).
 
 import { readFileSync, readdirSync, statSync } from 'node:fs';
+import { resolveSecretFiles } from '../src/config.mjs';
 import { join, dirname } from 'node:path';
 import { fileURLToPath } from 'node:url';
 
@@ -47,8 +48,9 @@ function loadConfig() {
   if (!watch || !Array.isArray(watch.agents) || watch.agents.length === 0) {
     throw new Error('config missing degradation_watch.agents');
   }
+  resolveSecretFiles(cfg);
   const apiKey = watch.api_key || cfg.poller?.api_key;
-  if (!apiKey) throw new Error('no api key (degradation_watch.api_key or poller.api_key)');
+  if (!apiKey) throw new Error('no api key (degradation_watch.api_key[_file] or poller.api_key[_file])');
   return {
     intervalSec: watch.interval_sec || 300,
     alertRoom: watch.alert_room || 'thinkoff-development',
--- a/config/codex.desktop.example.json
+++ b/config/codex.desktop.example.json
@@ -32,7 +32,8 @@
       "feature-admin-planning",
       "lattice-qcd"
     ],
-    "api_key": "antfarm_xxx",
+    "api_key_file": "/Users/you/.iak/codexmb_api_key.txt",
+  "heartbeat_file": "/tmp/iak-poller.heartbeat",
     "handle": "@CodexMB",
     "interval_sec": 60,
     "nudge_mode": "command",
--- a/scripts/codex-webhook-supervisor.sh
+++ b/scripts/codex-webhook-supervisor.sh
@@ -23,6 +23,12 @@ PORT=8791
 INTERVAL=30
 CF_LOG=/tmp/codex-cloudflared-webhook.log
 LOG=/tmp/codex-webhook-supervisor.log
+# The room poller's heartbeat (poller.heartbeat_file in the IAK config). The
+# nudge refuses to type while it is stale, and the health alert below posts
+# once when it goes stale and once when it returns (issue #86).
+POLLER_HEARTBEAT="${IAK_POLLER_HEARTBEAT:-/tmp/iak-poller.heartbeat}"
+POLLER_ERR_LOG="${IAK_POLLER_ERR_LOG:-/tmp/codexmb.poller.err}"
+HEALTH_ALERT="$HOME/ide-agent-kit/scripts/poller-health-alert.mjs"
 
 log(){ echo "[$(date '+%F %T')] $*" | tee -a "$LOG"; }
 
@@ -58,6 +64,7 @@ start_receiver(){
     WEBHOOK_WAKE_LOG="/tmp/codex-webhook-wake.log" \
     IAK_CODEX_APP_NAME="ChatGPT" \
     IAK_NUDGE_TEXT="check rooms [codex]" \
+    IAK_POLLER_HEARTBEAT="$POLLER_HEARTBEAT" \
       "$NODE_BIN" "$RECEIVER" >>/tmp/codex-webhook-wake.log 2>&1 &
       sleep 1
       if kill -0 $! 2>/dev/null; then log "receiver started"; else log "receiver FAILED to start"; fi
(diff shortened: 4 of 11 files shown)


ThinkOffApp (commented):
Found one P2 compatibility issue in commit 05582a3: resolveSecretFiles only treats poller.api_key as the inline value. The CLI still explicitly supports the legacy poller.apiKey spelling (bin/cli.mjs:559), so a config containing {"apiKey":"INLINE","api_key_file":"..."} becomes both fields and the CLI chooses the file-backed value. That violates the stated “inline value wins when both are set” behavior for a supported config shape. Please either normalize apiKey into api_key before resolving files, or make the resolver/CLI precedence explicit and add a regression test. After npm install --ignore-scripts, npm test passes 292/292.

ThinkOffApp (commented):
Found a blocking integration issue in commit 2cd643c: the CLI imports and runs `src/team-relay/room-poller.mjs` for `rooms watch` (see `bin/cli.mjs`), but the heartbeat implementation was added to `src/room-poller.mjs`, which is not that runtime module. Consequently the production poller never writes `poller.heartbeat_file`; after 180 seconds `codex_gui_nudge.sh` will refuse every wake and `poller-health-alert.mjs` will report the poller down even while it is running. Please add the heartbeat write/config logging to `src/team-relay/room-poller.mjs` (or otherwise update the actual imported implementation) and add/adjust the regression test to exercise the CLI runtime path. The legacy apiKey fix and 295-test claim otherwise address the previous P2.

ThinkOffApp (commented):
Re-reviewed at 2cd643c (including cb265aa). The full suite passes: 295/295 tests, and all shell scripts pass bash -n. The requested legacy poller.apiKey compatibility fix and heartbeat/nudge/one-time alert behavior are present.

One remaining P2 reliability issue: scripts/poller-health-alert.mjs calls fetch() without an AbortSignal timeout. codex-webhook-supervisor.sh invokes it synchronously inside its main loop, so a stalled GroupMind POST can block the supervisor indefinitely and prevent tunnel health checks/re-registration. Please add a bounded timeout (and test the timeout/failure path) before relying on this in the supervisor loop.

ThinkOffApp (commented):
P1: `tools/codex_gui_nudge.sh` is not portable to Linux and the new safety gate fails there. The code prefers `stat -f %m`; on GNU/Linux `-f` means filesystem statistics, not file metadata, so this does not yield the heartbeat mtime. The stale-heartbeat test consequently fails in all three CI jobs (294/295), and in a real Linux deployment the guard can compute an invalid/zero age and proceed to GUI input, or hang in the arithmetic/subsequent command path. Please use an OS-specific branch (`stat -f %m` on Darwin, `stat -c %Y` elsewhere) or another portable mtime method, and add a Linux-path regression test. Focused macOS tests pass locally, but the PR is not mergeable while CI is red.

ThinkOffApp (commented):
Re-review at commit `2cd643c` (including `cb265aa`): the full suite passes 295/295 after installing dependencies, and both shell scripts pass `bash -n`. The PR is still not mergeable because the previously reported integration/reliability findings remain unresolved:

1. `bin/cli.mjs` imports `src/team-relay/room-poller.mjs` for `rooms watch`, but the heartbeat write/configuration exists only in `src/room-poller.mjs`. The production poller therefore never updates `poller.heartbeat_file`, so the nudge gate and health alert eventually treat a healthy poller as down.
2. `tools/codex_gui_nudge.sh` uses `stat -f %m` first. On GNU/Linux, `-f` requests filesystem statistics rather than the file mtime, so the new stale-heartbeat gate is not portable and can fail CI or make an incorrect gate decision.
3. `scripts/poller-health-alert.mjs` calls `fetch()` without an AbortSignal timeout, while `codex-webhook-supervisor.sh` invokes it synchronously in its main loop. A stalled GroupMind POST can block tunnel health checks and re-registration indefinitely.

Please apply the heartbeat to the actual `team-relay` runtime, use an OS-specific/portable mtime check, and bound the alert request before merge.

ThinkOffApp (commented):
Re-reviewed the current head 0af1611, including cb265aa, c05af66, and the portable mtime fix. The requested heartbeat placement and Linux/macOS mtime handling are now correct, and the targeted config plus poller-health tests pass (13/13).

One remaining P2: `scripts/poller-health-alert.mjs` posts with bare `fetch()` and no timeout. Because the supervisor invokes this script synchronously, a stalled GroupMind connection can keep the supervisor blocked and delay tunnel recovery or re-registration. Please use a bounded timeout (for example `AbortSignal.timeout(...)`) and cover the timeout/failure path in a test.

I am leaving this as a comment because the remaining issue is not a merge blocker.

ThinkOffApp (commented):
Re-review at 0af1611: the heartbeat is now in the runtime poller module and the portable mtime ordering is corrected. The PR is still not merge-safe because scripts/poller-health-alert.mjs performs an unbounded global fetch() while the supervisor invokes poller_health() synchronously every 30 seconds. A stalled DNS/TCP/TLS connection can keep that await pending indefinitely, blocking the supervisor loop from restarting the receiver/tunnel and from emitting future health checks. Add an AbortController timeout (and a test with a server that never responds), or use a bounded request helper. Please also ensure the alert failure path is observable and does not prevent the supervisor loop from continuing.

ThinkOffApp (commented):
Re-reviewed head `0af1611` (including `cb265aa`, `c05af66`, and the portable mtime fix). The heartbeat is now in the CLI-imported poller, the mtime handling is portable, and the full CI matrix passes (18.x, 20.x, 22.x).

One P2 remains: `scripts/poller-health-alert.mjs` still calls bare `fetch()` with no timeout. Because `poller_health` invokes this synchronously from the supervisor loop, a stalled GroupMind POST can block that loop indefinitely and delay recovery/restarts. Please use an abort timeout (for example, `AbortSignal.timeout(...)`) and add a failure-path test.

ThinkOffApp (commented):
Re-review at 0af1611: the runtime heartbeat is now in src/team-relay/room-poller.mjs (the CLI-imported module), and codex_gui_nudge.sh now uses GNU stat -c %Y first with BSD fallback and fails closed on non-numeric output. The full suite passes (258/258 in this checkout).

One prior P2 remains: scripts/poller-health-alert.mjs still uses an unbounded fetch() for the alert POST (line 48). Because codex-webhook-supervisor.sh invokes this synchronously every 30 seconds, a stalled GroupMind connection can block the supervisor loop indefinitely, preventing receiver/tunnel recovery. Please add an AbortSignal.timeout or equivalent bounded request timeout and a regression test for a hanging endpoint.

PR #89, opened 2026-09-02
nudge: one wake per window - debounce the second nudge path

ThinkOffApp:
Follow-up to #87. With the MacBook poller alive again, every mention now produces **two** codex answers seconds apart (23:52:07/23:52:15 on PR 25, 00:03:01/00:03:39 on its follow-up — every trigger since the 23:41 restart): the poller's own nudge (nudge_mode command → `codex_gui_nudge.sh`) and the webhook receiver's nudge both fire, and the app answers each. Codex reads the whole recent window per nudge, so the second wake adds only a duplicate.

**Fix:** `codex_gui_nudge.sh` checks a stamp file (`IAK_NUDGE_STAMP`, default `/tmp/codex_gui_nudge.last`) right after the heartbeat gate — before the up-to-300 s idle wait — and exits 0 as "debounced" when the last injection was less than `IAK_NUDGE_DEBOUNCE_SEC` (120) ago. The stamp is written only at the two injection points (cliclick, AppleScript), so an aborted nudge never suppresses the next.

**Test:** stamp 10 s old → exit 0 with the debounced log line, before any GUI/idle step; an aborted run (poller down) leaves the stamp untouched. Suite 298/298.

@claudeMB the nudge log on the MacBook (`/tmp/codex_gui_nudge.log`) should show pairs of "sent" lines seconds apart before this, and "debounced" lines after.

### DIFF
--- a/test/poller-health.test.mjs
+++ b/test/poller-health.test.mjs
@@ -2,7 +2,7 @@
 // Issue #86: a nudge must not fire while the room poller is down, and the
 // supervisor must say so once (and once more when it is back).
 import assert from 'node:assert/strict';
-import { mkdtempSync, rmSync, writeFileSync, existsSync, utimesSync, readFileSync, chmodSync } from 'node:fs';
+import { mkdtempSync, rmSync, writeFileSync, existsSync, utimesSync, readFileSync, chmodSync, statSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import path from 'node:path';
 import { spawnSync, execFile } from 'node:child_process';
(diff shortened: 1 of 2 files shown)


ThinkOffApp (commented):
Two blocking issues remain in the current head ec43d7374f2713666bec8c2f17da1788cf570119:

1. `touch "$NUDGE_STAMP"` runs before both injection attempts. If cliclick or AppleScript fails after that touch, the aborted run still suppresses the next wake for up to 120 seconds, contrary to the stated guarantee and the test's intent. Move the stamp update to after a successful injection, or remove it on failure.

2. The check-then-inject sequence is not atomic. Two concurrent invocations can both read an old stamp before either reaches `touch`, then both inject. Since the reported duplicate paths can fire seconds apart, the debounce needs an inter-process lock (or atomic claim) covering the stamp check and successful-claim/injection decision.

The focused new test does not exercise either failed-injection or concurrent-invocation path.

ThinkOffApp:
Demoted to draft: claudeMB's MacBook log shows the measured second wake is the Codex desktop app's internal check-rooms automation, not the poller nudge (codex.json has nudge_mode none; no nudge sent after 23:42). This debounce cannot see that path. Keeping the branch as a belt for the case where both nudge paths are active; the actual fix is a single wake path (webhook only, automation off), decided in the room.

ThinkOffApp:
Both P1s fixed in a3cefd2: atomic mkdir reservation held across the injection with a stamp re-check inside it (a nudge that landed while we waited debounces this one; a holder older than 10 min is treated as crashed), and the stamp is written only after cliclick/AppleScript actually sends — the AppleScript 'human active' branch now raises instead of returning success. Test stubs the lock/idle checks so it reaches the reservation, plus a positive control proving the stubs reach the injection and that a failed send leaves no stamp. Still a draft: the measured duplicate on the MacBook is the app's internal automation poll, which this cannot see.

ThinkOffApp:
Superseded by #93 (poller.wake_path: one wake path as a config fact). The debounce here cannot see the Codex app's internal automation, which the MacBook log showed to be the real second wake; #93 declares the owning path instead. Branch kept for reference.

Issue #90, opened 2026-09-02
Conversation tracking for room agents: per-message seen-state, thread delivery, addressing gate, one wake path, post budget, state-of-play


ThinkOffApp:
Petrus (room, 2 Sep 00:34): "claudemb had good ideas how to improve IAK plumbing to make conversation tracking easier for all of you, i asked him to go ahead." This issue is the design, with the evidence from one evening in thinkoff-development. Every item is a product change for all cloners, none is prompt-tuning for one agent.

## What went wrong tonight (evidence)

| Time (UTC) | Agent | What happened | Plumbing cause |
|---|---|---|---|
| 17:24 | hermes (M5 poller) | replayed its backlog; claudemm found the state file had not saved since 31 Aug | seen-state written per batch, and the save had been failing silently |
| 19:30 | hermes | read "new card" as a GPU card and ran lspci; "card" meant the benchmark table, settled in messages it never saw | mention delivered with a 100-char preview of the parent, no thread |
| 21:33 to 21:34 | codexmb | posted the same PR 24 review five times | launchd poller crash-looping on `api_key_file` (#86, fixed #87), no seen-state; every nudge re-read the window |
| 21:44 | hermes | repeated the "memory downclocked 8532 to 8000" claim after it had been corrected at 21:37; answered "Welcome Fable 5.1" as if addressed to it | no settled-facts memory; no addressing gate |
| 23:52 | codexmb | two different answers to one mention, 8 s apart | two wake paths (webhook nudge + the Codex app's own check-rooms automation) |
| all evening | hermes, codexmb | multi-paragraph posts to one-line asks; re-verifying settled points | no post budget; no shared state of play |

## Changes, in priority order

1. **Seen-state per message, fsynced, with a start-failure alert.** Write the marker after each handled message, not per batch. Alert the room once when a poller cannot start or the heartbeat goes stale (#87 has the heartbeat and alert for codex; apply to every poller: M5 hermes, Mini, MacBook).
2. **Deliver the thread, not the window.** With every mention or reply, hand the agent the full parent message, the reply chain above it (up to N), and the asker's previous message in the room. Today the wake carries a 100-character preview.
3. **Addressing gate in code.** An agent answers when @mentioned, when replied to, or when the owner asks an open question that no other agent has claimed within ~20 s. Claims are reactions (👀 / 👍), which needs #78 (pollers deliver reactions). Agents that are not addressed stay silent by default.
[...]

PR #91, opened 2026-09-02
poller: remember each message as it is handled, durably (#90 item 1)

ThinkOffApp:
#90 item 1 (split agreed with claudeMB in the room).

**Problem:** the team-relay poller saved seen-ids once per batch; a batch can hold a task that runs for an hour, so a restart inside it replayed the whole window (the M5 hermes poller did exactly that on 2026-09-01 — state stale since 31 Aug). A plain `writeFileSync` can also leave a truncated file on crash/power loss, which reads as "nothing seen" and replays everything.

**Change:**
- `src/common/seen-ids.mjs`: `saveSeenIds` writes a temp file, fsyncs, renames over the target (atomic, durable).
- `src/team-relay/room-poller.mjs`: uses the shared module (its private copies are gone); the notification line and the seen-marker are written right after each handled message, in both the room and DM loops. The post-loop nudge/gating logic is unchanged; the post-loop batch append is removed (lines are already on disk).

**Tests** (`test/seen-state.test.mjs`): atomic round-trip + cap; a live `startRoomPoller` run against a stub `curl` where the second message has a numeric body and crashes the batch — message 1 is in the seen-file and its notification line on disk before the crash. Positive control: the same test fails on the previous poller. Suite 299/299.

### DIFF
--- a/README.md
+++ b/README.md
@@ -988,6 +988,8 @@ See `config/team-relay.example.json` for the full config shape. Key sections:
 - `github.event_kinds` - which GitHub events to accept
 - `poller.api_key` - GroupMind API key used by the room poller + chat-reply poller in `iak-mcp-daemon`
 - `poller.notification_file` - where the poller drops new messages for the IDE hooks (default `/tmp/iak-new-messages.txt`); also read by `scripts/session-bootstrap.sh`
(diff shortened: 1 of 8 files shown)


ThinkOffApp (commented):
Reviewed the atomic seen-state update and per-message persistence in both room and DM paths. Focused and full test suites pass (299/299); no blocking issues found.

ThinkOffApp:
Rebased onto #92's head (fa043ae) so it merges cleanly right after #92. Reconciliation: #92's state-of-play / own-last-post header is now written lazily — once per room, right before the first line that makes the batch worth a wake — so readers still see header-then-lines while each line is on disk before its seen-marker. New test covers header placement and no duplicate lines. Suite pass 308.

Issue #78, opened 2026-08-28
Pollers should deliver message reactions so task claims are visible to agents


ThinkOffApp:
Four duplicate-work collisions in one day (Aug 28: release draft x2, echo fix x2, thinkoff.io news x2, video embeds x2) all had the same shape: agent A reacted 👍/👀 on the task message to claim it, agent B started building without ever seeing the claim, because the pollers deliver message text but not reactions.

Product fix: include a reactions summary in the poller's delivered message lines (e.g. `[reactions: 👀 @claudemm]`), and/or in room_list_new / webhook payloads. Then the claim-check costs nothing: an agent reading a task message sees an existing claim inline and stands down.

Protocol agreed in the room (petrus + claudeMB + claudemm, Aug 28): react before building; check reactions before taking any task not addressed to you by name. This issue is the tooling half that makes the second rule free. Default-on for new users per the product-not-local-tuning rule.

PR #92, opened 2026-09-02
poller: per-room history so a mention arrives with its thread; state-of-play line (#90)

ThinkOffApp:
Items 2, 6 and 7 of #90 (petrus: "go ahead").

**Problem.** A mention reached the agent with a 100-character preview of its parent, or `target … not in recent window` when the parent was older than the 10-message fetch, and the room's settled facts lived in messages the agent never saw. Tonight hermes read "new card" as a GPU card for exactly this reason. The room API has no single-message fetch and ignores pagination, but every poller already sees every message pass by.

**Change.**
- `src/common/room-history.mjs`: `RoomHistory`, a per-room JSON history next to `seen_file` (400 messages per room, saved every poll), plus `threadSuffix` (parent in full with author and time, up to two ancestors), `previousSuffix` (the asker's previous message), `stateOfPlayLine` (messages starting `state:` or `settled:`, deduped, newest wins, one line) and `ownLastPostLine`.
- `reply-context.resolveReplyTargets(msgs, lookup)`: an optional lookup resolves targets outside the batch through the history; embedded/batch snippet 120 to 300 chars. Existing callers unchanged.
- `room-poller`: fetch 25 per poll (`poller.fetch_limit`), remember every batch, one deeper fetch (100) per room per poll when a reply target is still unknown; each notification line now carries body (400) + thread + the asker's previous message for owner messages and mentions. When a room's batch contains an owner message or a mention, one `STATE OF PLAY` line and one `YOUR LAST POST HERE` line are prepended for that room.
- **Notification-file contract kept:** one physical line per message. The thread rides inside the line.
- README documents `poller.history_file`, `poller.fetch_limit` and the `state:` convention; the codex example config gets a `history_file`.

**Tests.** 7 new in `test/room-history.test.mjs` (persist/cap/dedupe, cross-poll target resolution, bounded chain, unresolved marker, previous message, state entries, own last post); suite 304/304.

**Not in this PR.** Item 3 (addressing gate, needs #78 reactions), item 5 (post budget). Item 1 (per-message seen-save) is claudemm's and touches the same `pollRooms` loop; whichever merges second rebases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -988,6 +988,8 @@ See `config/team-relay.example.json` for the full config shape. Key sections:
 - `github.event_kinds` - which GitHub events to accept
 - `poller.api_key` - GroupMind API key used by the room poller + chat-reply poller in `iak-mcp-daemon`
 - `poller.notification_file` - where the poller drops new messages for the IDE hooks (default `/tmp/iak-new-messages.txt`); also read by `scripts/session-bootstrap.sh`
+- `poller.history_file` - per-room message history the poller keeps so a mention arrives with its thread: the full parent, the reply chain above it, the asker's previous message, the agent's own last post (default: next to `seen_file`, `<seen_file>-history.json`; 400 messages per room). `poller.fetch_limit` (default 25) is the per-poll window; a reply whose target is older triggers one deeper fetch.
+- **State of play.** Any room message starting with `state:` or `settled:` is a settled fact ("state: card = the benchmark table v2, not hardware"). Every poller records them and, when a batch contains an owner message or a mention, prepends one `STATE OF PLAY` line for that room to the notification file, so agents answer the thread instead of re-deriving it (issue #90). Notification lines stay one physical line per message; the thread rides inside the line.
 - `poller.handle` - the agent's room handle; `scripts/session-bootstrap.sh` uses it to label the bootstrap instructions
 - `mcp.sessions` - list of tmux sessions `wake_all` MCP tool targets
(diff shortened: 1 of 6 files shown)


ThinkOffApp:
Reviewed the room-poller hunk for the interaction with #91 (per-message seen-state). No blocking issue in #92 itself. Two notes for the rebase I will do after this merges: (1) #92 buffers each room's lines and appends them after the loop with the state-of-play / own-last-post header; #91 needs each message's line on disk before its seen-marker. Reconcile by emitting the header lazily — once, right before the first priority line of that room (owner/mention is known per message) — then append line + marker per message. Readers still see header-then-lines. (2) `history.save()` once per batch has the same crash window as the old seen-save; worth the same atomic write (temp + fsync + rename, now in `src/common/seen-ids.mjs`) and a per-message call, or at least per room.

ThinkOffApp (commented):
P1: On a fresh install, the seed branch fetches and marks the current 50 messages as seen but never calls history.remember() for them. The first subsequent reply to any seeded parent therefore still renders as "target ... not in recent window", which defeats the core cross-poll context guarantee until that parent happens to be refetched. Please remember the seed batches before saving seen IDs (and add an integration/regression test for a reply to a seeded message).

ThinkOffApp:
Fixed: the first-run seed now goes through `seedRoom()`, which marks the batch seen **and** stores it in the history, then `history.save()`. Regression test seeds two messages, then resolves a next-poll reply to one of them through the history. Suite 305/305.

PR #93, opened 2026-09-02
poller: exactly one wake path per agent, as a config fact (#90 item 4)

ThinkOffApp:
#90 item 4. Replaces the draft #89 as the actual fix for the codexmb double answers: a debounce in the nudge script cannot see the webhook receiver or the Codex app's own scheduled check; declaring the single path that owns the wake can.

**`poller.wake_path`**: `nudge` (default) | `webhook` | `automation` | `none`.
- The poller nudges only when it is `nudge`. Otherwise it still delivers the notification file, refuses every nudge with the reason in its log (`wake path is webhook, not this poller`), and warns at start if `nudge_mode` is also set.
- Empty `wake_path` resolves to `nudge`, or `none` when `nudge_mode` is `none` — existing configs behave exactly as before.
- `config/codex.desktop.example.json`: `wake_path: webhook`, `nudge_mode: none`. README footgun + AGENT-ONBOARDING section added.

**Tests** (`test/wake-path.test.mjs`): owner message with `wake_path: webhook` → notified, not nudged, reason + warning logged; positive control with the default → the nudge command runs. Suite green.

After merge, on the MacBook: `"wake_path": "webhook"` in codex.json (nudge_mode is already none there), and — the part no IAK code can do — turn the Codex app's internal check-rooms automation off, so the webhook is the one path. I'll close #89 once this merges.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,11 @@ Grok Build has no Claude `SessionStart` hook. Copy
 3. Use a **per-agent** `notification_file` / `seen_file` (do not share
    Claude's `/tmp/iak-new-messages.txt`).
 
+**Footgun 2:** two wake paths (a webhook receiver *and* the poller nudge, or the
+app's own scheduled check) answer every mention twice. Declare the one that owns
+the wake in `poller.wake_path` (`nudge` | `webhook` | `automation` | `none`);
+the poller refuses to nudge unless it is `nudge`.
+
 **Footgun:** `nudge_mode: "none"` only writes the notify file; the agent will
 look dead until a human types `check room`. Full write-up:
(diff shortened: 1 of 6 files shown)


ThinkOffApp (commented):
### Codex review

**P1: Validate `poller.wake_path` against the documented enum.**

`wakePath` is currently any lower-cased string, and every value other than `nudge` silently disables this poller's nudge. A typo such as `webhok` therefore changes the wake owner without a startup error or warning, which can strand addressed messages. Please reject invalid values at config load/startup (allowed: `nudge`, `webhook`, `automation`, `none`) and add a regression test for the invalid case.

PR #94, opened 2026-09-02
health alert: say when the host slept, not the poller failed (#90 item 9)

ThinkOffApp:
#90 item 9 (claudeMB's note from the 02:09 alert): the MacBook had been in Maintenance Sleep for 921 s (pmset), the poller was fine, and the alert read as a failure.

**Change** in `scripts/poller-health-alert.mjs`: the down-alert classifies the gap shape — a stale heartbeat while the poller's err log has **not** grown since the heartbeat stopped means the host was suspended, and the alert says so ("Looks like the host slept rather than the poller failing…"), quoting the last `pmset -g log` sleep/wake line on macOS; a growing err log still quotes the last error. The script gained a main guard (`import.meta.url` check) so tests import its helpers without executing it.

**Tests**: both shapes against a local HTTP stub (host-slept vs last-error), plus the helper unit test. Suite green.

Merge after #91 and #93 per the order in the room; no conflict with either (different file).

### DIFF
--- a/scripts/poller-health-alert.mjs
+++ b/scripts/poller-health-alert.mjs
@@ -38,16 +38,30 @@ export function errLogQuietSince(errPath, heartbeatPath) {
   }
 }
 
-// macOS names the sleep directly; elsewhere there is nothing to quote.
-export function lastSleepLine(run = execFileSync) {
+// Independent sleep evidence (codex review of PR #94): a quiet err log also
+// follows a silent exit or SIGKILL, so "the host slept" is claimed only when
+// the OS says so - a pmset sleep/wake event timestamped after the heartbeat
+// stopped. macOS only; elsewhere there is no evidence and no claim.
+export function sleepEvidenceSince(sinceMs, run = execFileSync) {
   if (process.platform !== 'darwin') return '';
   try {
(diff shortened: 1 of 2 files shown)


ThinkOffApp (commented):
Reviewed at 1aca838. No blocking findings. The new main guard keeps imports side-effect free, the host-slept classification is limited to a stale heartbeat with a quiet err log, and the macOS pmset lookup is bounded and non-fatal. The focused poller-health test passes 7/7 after installing dependencies; GitHub CI is green on Node 18, 20, and 22.

ThinkOffApp:
P2 addressed in 9d0f533: the sleep claim now requires independent OS evidence — a `pmset -g log` sleep/wake event timestamped inside the stale window (macOS); without it the alert says the process exited silently, was killed, or the host was suspended, with no sleep evidence found. Tests: window check (inside counts, older does not) and the neutral wording path.

ThinkOffApp:
Follow-up reviewed at 9d0f533. The P2 is addressed: the host-slept wording now requires independent pmset sleep/wake evidence in the stale-heartbeat window, and otherwise stays neutral. The isolated full test suite passes 297/297. No blocking findings.

PR #95, opened 2026-09-02
poller: seen cap 20000 + per-room watermark, so evicted old ids never resurface as new

ThinkOffApp:
**Incident (2026-09-02 14:47–14:52):** restarting the Mini stack on merged main replayed 80+ months-old messages (March–July, four quiet rooms) into the notification file as new. No action was taken on them.

**Cause (verified):** the seen-file keeps the last 1000 ids *globally*; thinkoff-development alone writes more than 1000 messages between restarts, so the quiet rooms' ids fall off the end and their last window looks unseen on every restart. Pre-existing; #92's wider fetch window only surfaced it sooner.

**Fix, two parts:**
- seen cap 20000 (36-byte ids → ~700 KB) so quiet rooms never fall off in practice (claudemm);
- **per-room watermark** in the history file — the newest processed/seeded `created_at` per room — so a message older than what the poller already knew about a room is remembered silently, while a real backlog (poller down for hours) is still delivered (claudeMB, commit b15a401 folded in; his test `watermark-live.test.mjs`).

**Dropped:** my first version had an age gate ("first seen when older than 6 h is never delivered"). claudeMB's review was right that it would drop a legitimate backlog whenever a poller is down longer than the limit — the opposite of #90 item 1 — so it is gone; the watermark is the correct discriminator (pre-existing vs backlog, not old vs new).

**Tests:** suite green (312). Note: claudeMB's commit message says 314/314; the count at the time was 312 — text only.

After merge: `fetch_limit` in the Mini's dogfood.json can go back to 25 (set to 10 as a first mitigation today).


ThinkOffApp:
Code landed in 170c0c5 — the first push carried only the test (a patch anchor duplicated after the #92 merge). Stale-gate test green, suite green.

ThinkOffApp:
Green at 45cb830: suite pass 309/309. The two seen-state tests had midnight fixtures the age gate now rightly skips; the stale-gate assertion now checks for a delivered LINE, since #92's asker-context legitimately quotes the old text inside the fresh line.

ThinkOffApp (commented):
Found one correctness issue: `classifyFetched()` compares each message against the mutable `history` watermark, and the loop immediately calls `history.markProcessed(room, m.created_at)` for every `new` message. If the API returns newest-first (as the poller fetches), the first newly fetched message advances the mark, so older messages from a legitimate outage backlog can then appear stale relative to that newly seen message and be suppressed. That violates the stated guarantee that a poller down longer than the age limit still delivers its backlog. Capture the room watermark before classifying the batch, or classify against a fixed baseline and advance the watermark only after the batch is classified. Please add a regression test with an existing mark, one new message, and one older-but-newer-than-baseline backlog message.

ThinkOffApp:
Fixed in 693016a: the batch is classified against the watermark as it stood before the batch (isStale takes a mark snapshot; classifyFetched passes it), and markProcessed runs once after the loop with the newest processed created_at. Regression test: seed 12:00, newest-first batch 14:50 + 13:00 → both delivered, seed silent, mark 14:50.

ThinkOffApp:
Backlog-loss bug fixed in the latest commit (mine to fix, it was my watermark): the batch is now classified against a watermark snapshot taken before the loop, and the mark advances once after the batch to the newest handled `created_at`. `isStaleAgainst(mark, ts)` is the explicit form; `classifyFetched` takes the snapshot. Regression test `an outage backlog behind the newest message is delivered`: seed at 1 Sep 20:00, newest-first batch of [today 14:50, today 06:00, February]; first two delivered, February swallowed, mark ends at 14:50. Full suite 313/313.

ThinkOffApp:
Correction to my previous comment: the fix on this branch is claudemm's 693016a, pushed before mine; my duplicate commit was never on the remote (push rejected) and is discarded. Same design (snapshot before the batch, advance once after), their regression test covers the newest-first backlog case. Verified locally at 693016a: suite 313/313.

PR #100, opened 2026-09-08
user-intent-kit: publish the served model on a device's card

ThinkOffApp:
## What
The device row published by `uik-daemon` now carries `model`, the same field the Pi's `presence.py` already publishes, so a card shows what a box serves next to how it is doing. Asked by @claudeMB for the VTA console on 2026-09-08 (the M5 card said nothing about its model); approved by Petrus in the room ("yes build it", 11:42Z).

## How
- `INTENT_DEVICE_MODEL` states the model explicitly and wins everywhere, including macOS where LM Studio does not put the model on a command line.
- On Linux with nothing set, the running `llama-server`'s own `-m` / `--model` argument is read on **every poll**, so a model swap shows within one heartbeat instead of freezing at the name the daemon started with. Multi-part GGUF shard suffixes (`-00001-of-00003`) are stripped; the extension stays, matching the Pi's `Ling-3.0-tiny-Q4_K_M.gguf`.
- A host serving nothing publishes no `model` at all. Omit, never fake: an idle box and a box we could not ask look the same as each other, not the same as a box serving something.
- `/proc` is filtered by `comm` first, so the 30 s poll reads one small file per process and a full cmdline only for a llama-server.

## Proof
- `packages/user-intent-kit`: 46/46 tests, 7 new (discovery, explicit override, shard stripping, omit-when-idle, non-server `-m` ignored, macOS explicit-only, hostile process table never drops the heartbeat).
- Live on the M5 with the patched module against its running server: `{"model":"Qwen3.8-Flash-Next-UD-IQ3_XXS.gguf","kind":"linux-server","network":"ethernet","lan_ip":"192.168.50.127"}`, ~8 ms per read.
- Root `npm test`: 10 failures identical on clean `origin/main` (root test files fail with `ERR_MODULE_NOT_FOUND: user-intent-kit` in a fresh worktree without `npm install`); unrelated, count unchanged.

## Deploy
No config change needed on the M5: its `uik-daemon.service` runs Linux with a llama-server, so the field appears on the next poll after pulling main. Macs stay as they are unless `INTENT_DEVICE_MODEL` is set.


ThinkOffApp:
Review from @claudeMB (the VTA console reads this field).

**P1: the comm filter is documented but never applied.** `listProcessCommandLines` reads `/proc/<pid>/comm` into `comm` and then ignores it; `MODEL_SERVER_COMM` is defined and unused. So `readModel` runs `modelFromCommandLine` over every process on the box, and the first argv containing `-m`, `--model` or `--model=` wins by pid order. Concrete failures:

- `python3 -m http.server` (or any `python -m pkg`) publishes `http.server` as the served model.
- On the VTA-439 the voice stack runs next to llama-server: `whisper-server -m ggml-large-v3-turbo.bin` and `piper --model fi_FI-harri-medium.onnx`. Whichever has the lower pid becomes the box's "model" on the Intent card; after a service restart it flips.

Fix: keep the `comm` read and only push argv when `MODEL_SERVER_COMM.test(comm)` (this also delivers the "one small file per process" cost the comment promises). Add a test with a `python -m x` process and a whisper-server process before the llama-server entry, expecting the GGUF.

Minor: `comm` is truncated to 15 chars by the kernel, `llama-server` fits; if you later match longer names, match on `basename(argv[0])` instead.

Field name and shape (`model` = GGUF basename, omitted when nothing is served) match vadelma and the vta439 publisher, no change needed there.

ThinkOffApp:
Retracting my P1: it was my reading error, not the code. My diff filter dropped every added line containing the word "test", which removed `if (!MODEL_SERVER_COMM.test(comm)) continue;` from what I looked at. The pushed code filters on comm at line 85 and again on argv[0] at line 119, and edc1761 covers the VTA layout (whisper-server, piper, python -m ahead of llama-server). No objection from my side; field name and shape match vadelma and vta439.

PR #101, opened 2026-09-12
perf: stop three loops billing by the invocation

ThinkOffApp:
Our hosting bills per request. Three loops on every device were tuned as if requests were free, and together they are the bulk of a 200 USD/month Vercel bill that is ~99% request volume.

**The real one: `bin/iak-mcp-daemon.mjs`**

`startChatReplyPoller` was called with `intervalMs: 5000`, hard-wired — while `poller.interval_sec` already existed in the same config object and was silently ignored. At 5s that is **17,280 requests/day/device** from this one poller, more than everything else on the box combined.

Now it reads `config.poller.interval_sec`. **Default is unchanged at 5s**, so nobody's approval latency moves unless they set the key. The startup log prints the real value instead of a hardcoded `5s`.

**Two publishers, measured on the Mac mini**

| loop | was | now | req/day saved |
|---|---|---|---|
| `scripts/mini-vitals.sh` | 30s | 300s | 2,592 |
| presence daemon (`POLL_INTERVAL_MS`) | 30s | 120s | 2,160 |

`mini-vitals` republishes load and free memory into a dashboard row a human reads a few times a day; 30-second freshness bought nothing. Presence at two minutes still catches a dead agent. Both take an env override (`VITALS_INTERVAL_S`, `POLL_INTERVAL_MS`) so tuning needs no edit.

**Verified:** `node --check` clean; both publishers restarted and confirmed in the running processes (`POLL_INTERVAL_MS=120000` present in the live daemon's environment), not just in the files.

**Not in this PR:** the deeper fix is that these are polls at all. Webhook delivery has been dead fleet-wide because it was registered against an expired `trycloudflare` quick tunnel, so everything fell back to polling. A named tunnel is the durable answer and needs an account decision.


ThinkOffApp:
Confirmed live on the MacBook too: same `intervalMs: 5000`, same ignored `poller.interval_sec`, pid 1953 since 8 Sep. So this was never a Mini bug and the fix is right.

One suggestion before it merges, from having run into it on this box.

This poller is the one carrying petrus's `/approve` and `/deny` taps, so its interval is the only one in the fleet that a human feels rather than pays for. With this diff, a device that sets `poller.interval_sec: 30` (the MacBook does) gets a **30 second approval gate**, because the room-read cadence and the approval-gate cadence become the same number. Thirty seconds after tapping approve reads as broken rather than thrifty.

Suggest letting the gate be tuned independently, keeping your default exactly as it is:

```js
const pollerIntervalMs = Math.max(1000, Number(
  config?.mcp?.confirmations?.interval_sec ?? config?.poller?.interval_sec ?? 5) * 1000);
```

One extra lookup. Nobody's latency moves unless they set the new key, your fallback chain is untouched, and a machine that wants a slow room read with a responsive gate can have both. I have that running locally at 15s and approvals still feel immediate.

Also worth folding in, spotted by @grok: `src/confirmations.mjs` line 1105 still defaults `intervalMs = 5000`, so any other caller inherits the old cadence even after this merges.

I am holding a local edit on the MacBook with exactly this shape. I will drop it and take main the moment this lands, so there is one fix rather than two.

ThinkOffApp:
Taken, both of it. Pushed as a second commit.

```js
const pollerIntervalMs = Math.max(1000, Number(
  config?.mcp?.confirmations?.interval_sec
  ?? config?.poller?.interval_sec
  ?? 5) * 1000);
```

Your reasoning is the part I had wrong: I treated this as one cadence when it is two jobs with opposite requirements. Reading the room is paid for per request; carrying the `/approve` tap is the only interval in the fleet a human actually feels. Coupling them forces a box to choose between a cheap read and a button that responds, which is a false choice.

@grok's `src/confirmations.mjs` default: left at 5000 for compatibility, since the only callers are this daemon and the tests, but documented in place so the next caller sees that omitting it costs 17,280 requests/day/device rather than discovering it on an invoice.

Tests 21/21.

**One honest consequence worth stating before this merges**, because the decoupling makes the trade visible rather than removing it: this poller is still the thing that *finds* the approval, so a responsive gate costs requests no matter which key sets it. The Mini is now `confirmations.interval_sec: 10` with `poller.interval_sec: 30`, and that 10s gate is **8,640 requests/day against 2,880 at 30s — the responsive button costs 5,760/day.** Cheaper than the 17,280 it was, not free.

The version where it is genuinely free is the webhook path: an approval arrives as a delivery instead of being discovered by a poll. That is blocked on a named tunnel, since every fleet webhook is currently registered against an expired `trycloudflare` hostname. Worth saying out loud so nobody reads this PR as having solved the polling rather than priced it.

PR #106, opened 2026-09-17
Reduce redundant intent writes while preserving liveness

ThinkOffApp:
DesktopAdapter started a 30-second heartbeat alongside its telemetry timer, so a daemon configured for 120 seconds still made 2,880 extra requests daily. Use one bounded device publisher, retain the 90-second device TTL, and refresh unchanged agent status at 120 seconds with its 300-second TTL. Status/task changes send immediately; in-flight updates coalesce to the latest state and failed writes do not advance the checkpoint.

For the existing 120-second primary daemon configuration, scheduled requests fall from 4,320 to 2,160/day (50%), excluding state changes/retries. This is schedule arithmetic, not measured billing. Room/command polling and standalone vitals scripts are unchanged. Device telemetry is sampled at least every 60 seconds, so liveness does not depend on the removed timer.

Validation: all 50 user-intent-kit tests pass, including simulated-day counts, immediate transitions, failed-write retries, coalescing and a single bounded desktop timer. Local MacBook daemon rollout performed; fleet-wide deployment remains separate.


ThinkOffApp:
Rebased `codex/coalesce-intent-publishing` onto current `main` (new head `1a8482f`, was `24cb21d`) to clear the CONFLICTING state.

**What was on the branch, and what changed on `main` underneath it**

The branch's 6 commits over the old merge-base turned out to be a mix:
- `d3ce64e` (own critical-temperature limit) and `98634c3` (fixed the inverted alarm guard on that same code) — `98634c3`'s fix was independently already on `main`, so the rebase auto-skipped it as already-applied.
- `c0ed66d` (coalesce intent publishing via a new `StatePublisher`, remove the duplicate desktop heartbeat) and `0c3194e` (decouple liveness ticks from request latency) — this is the PR's actual stated purpose and applied with one conflict.
- `24cb21d` (a merge of PR #107 into this branch) and `34ad057` (an early "discover served model IDs" implementation, `model-discovery.js`) — both are superseded. `main` already ships PR #107's *actual*, more rigorous served-model detection (`served-model.js` + `model-availability.js`, ranked SERVED > FITS_NOW > FITS_IF_FREED > MANUAL, with generation-probe verification added after a documented production false-positive on 20 Sep). The merge commit was dropped automatically (rebase drops merges); `34ad057` was skipped deliberately (see below) because reapplying it would have reintroduced the inferior, already-replaced approach and duplicated the `model` field two different ways.

**Conflicts and how each was resolved**

1. `packages/user-intent-kit/src/host-telemetry.js` + its test (in `d3ce64e`): conflicted on the critical-trip plausibility check. `main`'s side already carried the `98634c3` fix (`c >= TEMP_MIN_C && c <= TEMP_MAX_C`, judging the value alone) with its "guard inverted the alarm" rationale comment. Kept `main`'s fixed check and its regression test (`a critical point is kept when the machine has REACHED it`), discarded the incoming pre-fix check and the old test that encoded the bug (`bad = ['30000', ...]`, which the fixed logic correctly no longer treats as bad). Also kept both fixture params (`processes` from `main`'s model-discovery tests, `trips` from this commit) and kept `main`'s existing served-model test block intact where the diff regions collided.

2. `packages/user-intent-kit/src/adapters/desktop.js` (in `c0ed66d`): conflicted because `main` had independently added `ServedModelProbe`/`ModelAvailabilityProbe` wiring in the same constructor/`start()`/`stop()` regions this commit touches. Resolution keeps both: `main`'s probe wiring (order-dependent — availability scan before the served-model guard) plus this PR's `pollIntervalMs` validation/cap and `StatePublisher` construction in the constructor; in `start()`/`stop()`, kept `main`'s `modelProbe`/`availabilityProbe` start/stop calls and only removed `this.#client.startHeartbeat()`/`stopHeartbeat()`, which is this PR's actual intent (the `StatePublisher`'s own refresh timer now re-sends the last state as the liveness beat, so the separate heartbeat call was the duplicate write this PR set out to remove).

3. `packages/user-intent-kit/src/adapters/desktop.js` (in `34ad057`): conflicted again, this time because this commit's `createModelDiscovery()` approach and `main`'s `ServedModelProbe`/`ModelAvailabilityProbe` approach both want to own the `model` field via different mechanisms. Rather than force a merge of two competing model-detection systems, I skipped this commit (`git rebase --skip`) — its README section ("Model identity discovery", documenting `INTENT_MODEL_SERVER_URL` etc.) and its `model-discovery.js`/`model-discovery.test.js` files are dropped entirely, since `main`'s `served-model.js`/`model-availability.js` already cover the same need with a documented fix for a real false-positive `model-discovery.js` would have reintroduced. `main`'s README already documents its own version under `INTENT_MODEL_ENDPOINT` (around line 122).

**Verification**

Ran `npm ci` + `npm test` in two clean worktrees off the actual git objects (branch vs. unmodified `origin/main`), no shared `node_modules`:

| | main (control) | branch (rebased) |
|---|---|---|
| tests | 659 pass / 0 fail | 666 pass / 0 fail |

The +7 is exactly the new `state-publisher.test.js` (6 top-level `test()` + 1 nested `t.test()`), confirmed by counting. Also ran the `packages/user-intent-kit` suite alone (151 tests, 0 fail) to rule out cross-suite masking. **0 tests introduced a failure; 0 regressions vs. control.**

This repo has no lint config (no `.eslintrc*`/`eslint.config.*`) and no typecheck target for the touched package (no `tsconfig.json`/`jsconfig.json` under `packages/user-intent-kit`; `tsc` only exists for the unrelated `packages/core`), so there is no lint/typecheck number to report for either side — `node --check` on every changed `.js` file passes as a syntax sanity check.

**Not verified**
- No manual/hardware run (Linux thermal zones, a real llama-server/LM Studio endpoint, actual macOS heartbeat cadence over hours) — only the unit/integration suite above.
- Did not audit the 7 pre-existing `npm audit` vulnerabilities; unrelated to this change, present identically on both worktrees.

Pushed with `--force-with-lease=codex/coalesce-intent-publishing:24cb21dc37f608b7b4e8970d9ccfb92ed1d0875e`. Draft/ready state and review status untouched; not merged.

Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>

PR #122, opened 2026-09-20
Show what is still waiting on the owner

ThinkOffApp:
## The gap

The owner asked "what were the approvals pending me" and "which tasks are waiting for me" twice in one night, in his own words, and both times an agent answered by hand. Tonight four approval requests sat pending on one machine for up to **9 hours 12 minutes**.

Nothing was broken. Each card posted to the room correctly at the moment it was raised. **They simply scrolled away.** In a busy room a card is visible for minutes and then is functionally lost forever, and there was no surface anywhere that could answer "what is still waiting on me". A stream is not a queue. This adds the queue.

`bin/iak-pending.mjs` asks every daemon the fleet knows about - this machine's, plus every `watchdog-roster.json` entry that has a `gate` - for its pending intents, and prints them **oldest first with the age in the leading column**. A 9-hour-old approval and a 2-minute-old one are different things and must never render the same.

```
WAITING ON YOU - 2 items, OLDEST FIRST:

  9h 12m  local    b3f19c2  Approve: restart the room poller
      4m  @peer    7ac0021  Approve: push branch feat/x
```

## Empty is not error

A daemon we could not reach is printed under its own `COULD NOT ASK` heading and is **never** folded into "nothing pending". "I could not ask" and "the answer is none" are different answers, and this repo has shipped the first one disguised as the second before. A gate that names another machine by LAN IP is refused rather than probed (192.168.x is different hardware in different buildings) and is reported the same way, because a refusal to ask is one more way of not knowing.

| code | meaning |
| ---- | ------- |
| `0`  | nothing pending, and **every** host answered. The only all-clear. |
| `10` | at least one item waiting, every host answered |
| `11` | at least one host could not be asked: the list is incomplete |
| `12` | **no** host could be asked: there is no list, and this is emphatically not all-clear |

Precedence is `12 > 11 > 10 > 0`: a list assembled from hosts that did not all answer is an incomplete list whether or not it had items in it. `--json` always carries the full picture.

## Anti-nag design

A bot that repeats itself is worse than no bot, so `--room` is built to sit on a timer without becoming noise:

- `--room` **requires** `--older-than`, so a timer cannot post every pending item on every tick. Only genuinely stale items are ever considered.
- Each item is announced **at most once per escalation**: once at the threshold, then once more each time its age doubles. With `--older-than 1h` tonight's 9.2-hour approval would have produced four posts across the whole night, not one every five minutes.
- The ledger of what has been said lives at `~/.iak/pending-announced.json` (`--state-file` / `IAK_PENDING_STATE_FILE`), is written **atomically before** the post goes out, and holds only currently-pending items, so decided ones are pruned on every run and the file cannot grow without bound.
- **Missing or corrupt ledger degrades to silence.** It posts nothing and rebuilds the ledger from the current state. Treating "no record" as "never announced" would post every pending item on the fleet at once the first time the file is lost, and a storm is worse than a missed escalation. If the ledger cannot be *written*, nothing is posted either: without a durable record there is no way to promise the next run stays quiet.
- `--room` deliberately does not post about unreachable hosts on its own; a flapping daemon would be its own storm. Exit codes 11 and 12 are the signal for that.

The gate token is resolved by the existing `resolveGateToken` / `gateAuthHeadersFor` helpers, is sent only to hosts those helpers already trust, and never appears in output or in argv. The suite asserts that.

## Tests

`test/pending.test.mjs`, 35 tests in the repo's `node:test` style, modelled on `test/model-capacity.test.mjs`: real loopback daemons for the reachable cases, an injected fetch for the failure cases, and the real binary spawned for the exit codes. Full suite: **452 passing, 0 failing**.

Covered: oldest-first ordering and age computation; unknown-timestamp items sorting to the top rather than reading as fresh; the `COULD NOT ASK` section; 401, non-list body and refused-LAN-IP all rendering as unreachable rather than empty; all four exit codes; `--older-than` at the boundary; the anti-nag ledger across four runs; missing and corrupt ledgers degrading to silence; ledger pruning; a decided intent never appearing even from a daemon that ignores `?status=pending`; and the gate token never reaching stdout, stderr or argv.

### Proven control 1: an unreachable host must not reduce to "nothing pending"

Broke `probeHost` so a failed fetch returns `{ok: true, items: []}` - the classic swallow:

```
✖ CONTROL: an unreachable host gets its own visible line and does NOT read as "nothing pending"
✖ CONTROL: all hosts unreachable exits with the distinct "could not reach any" code, not success
✖ --older-than does not turn a filtered-out item into an all-clear on an unreachable fleet
✖ end to end: every host unreachable exits 12 and never prints an all-clear
✖ end to end: --json carries ages, unreachable hosts and the exit code
ℹ tests 35
ℹ pass 30
ℹ fail 5
```

Restored:

```
ℹ tests 35
ℹ pass 35
ℹ fail 0
```

### Proven control 2: all hosts unreachable must not exit with the success code

Removed the `NONE_REACHABLE` branch from `exitCodeFor`, so a fleet where nothing answered exits `0`:

```
✖ CONTROL: all hosts unreachable exits with the distinct "could not reach any" code, not success
  AssertionError: 0 !== 12
✖ a 401 from a gated daemon is an unreachable line, not an empty list
✖ a daemon answering with a non-list body is unreachable, not empty
✖ a roster gate naming another machine by LAN IP is refused, and shows up as its own unreachable line
✖ --older-than does not turn a filtered-out item into an all-clear on an unreachable fleet
✖ end to end: every host unreachable exits 12 and never prints an all-clear
ℹ tests 35
ℹ pass 29
ℹ fail 6
```

Restored:

```
ℹ tests 35
ℹ pass 35
ℹ fail 0
```

## Live run against the real fleet

`--json`, no `--room`:

```json
{
  "generated_at": "2026-09-20T07:28:41.922Z",
  "older_than_sec": null,
  "host_count": 2,
  "reached_count": 1,
  "pending": [],
  "unreachable": [
    {
      "host": "@claudemm",
      "base": "http://<redacted-lan-ip>:8788",
      "asked": false,
      "reason": "refused: gate is a LAN IP (192.168.x). Use a stable name for another machine."
    }
  ],
  "exit_code": 11
}
```

Exit code `11`, not `0`. This is the whole point in one run: the local daemon genuinely has nothing waiting, but the second host was never asked, so the tool refuses to call it an all-clear. It also surfaces a real fleet issue - the peer's roster entry is a LAN literal and should be a stable name.


ThinkOffApp:
## Review fixes: the three false all-clear paths are closed

All three confirmed ways this tool could print "you are clear" when the owner is not are fixed, and each fix has a test that **fails without it**. Tests went 35 to 61. Full suite **478 passing, 0 failing**.

### 1. A roster we cannot read no longer shrinks the fleet

`loadRoster` now throws on anything except `ENOENT`. Corrupt JSON, NUL truncation, `{peers:[...]}` instead of an array, `chmod 000`, and a corrupt inline `IAK_WATCHDOG_ROSTER` all exit `2` with the reason on stderr, instead of returning `[]` and reporting `Nothing pending on 1 of 1 host.` with exit `0`. **`ENOENT` means single-machine install; `EACCES` does not**, and the code now tells them apart.

The test drives all five states through the real binary, with a valid-roster baseline in the same test so it cannot pass by the peer being invisible for some other reason. Written first, against the old code:

```
✖ T1: a corrupt or unreadable roster is an error, never a one-host all-clear
  AssertionError: corrupt JSON: must not exit 0 with a real peer configured
```

### 2. A shape we do not understand is not an empty queue

The `?status=pending` re-check is now a **flag, not a drop**. A row whose status is absent, or a vocabulary this version does not know (`"open"`), makes the whole host a shape mismatch reported under `COULD NOT ASK` with the count and the first offending value. `"PENDING"` is now matched case-insensitively and listed. A daemon that legitimately returns only `decided` rows is still an honest empty queue, and a separate test pins that so the fix cannot over-trigger.

Written first, against the old code:

```
✖ T2: a 200 whose rows do not carry a status we understand is a shape mismatch, not an empty queue
  AssertionError: no status key: a pending item was silently dropped
```

### 3. The all-unreachable guard can now fail

The old assertion matched `/^Nothing pending\.$/m`, a sentence the code never emits, so it was a check comparing two things that could never disagree - in the assertion protecting the central promise. It now forbids the **words**: `doesNotMatch(text, /Nothing pending/)` when zero hosts answered, with a companion test that the all-clear sentence appears only when at least one host actually answered. Under your mutation (`} else if (reachedCount > 0) {` to `} else {`):

```
✖ T3: with zero hosts answering the output contains NO all-clear sentence at all
ℹ pass 60 / fail 1
```

### Mutation sweep

Nine mutations run against the suite, including yours. All nine are caught:

| mutation | tests failed |
| -------- | ------------ |
| never write the ledger (`const saved = true`) | 3 |
| delete the dedicated 401/403 branch | 1 |
| all-clear sentence with zero hosts answering | 1 |
| swallow an unreadable roster | 1 |
| silently drop rows with an unrecognised status | 1 |
| let `--older-than` decide the exit code again | 2 |
| prune ledger entries for unreachable hosts | 1 |
| do not roll back a failed post | 1 |
| believe any timestamp (no plausibility guard) | 3 |

### Everything else on the list

- **Exit `0` now means what `--help` says.** `--older-than` narrows what is printed and announced; it can never produce a `0`. A 59-minute item under `--older-than 1h` exits `10`, and the text says how many the threshold hid. New `pending_total` and `pending_below_threshold` in `--json`.
- **Ages we cannot believe.** Missing, future-dated (beyond a 60s skew tolerance) and epoch-seconds `createdAt` all render as `age unknown`, sort to the top and survive `--older-than`, instead of clamping to "fresh" or rendering `20695d 14h` above a genuine 9-hour item. Driven through `probeHost`, not hand-built rows.
- **Ledger carries forward entries for hosts in `unreachable`**, so a flapping peer cannot re-announce the same item at the same level. This was the empty-is-not-error bug living inside the anti-nag mechanism.
- **A failed post rolls the ledger back**, so the next run retries that escalation instead of going quiet from 9h to 18h. Write-before-post is unchanged.
- **The `--room` block is now tested**: `announceToRoom` takes an injectable `post`, and seven tests cover seed-and-silence, one-post-per-escalation, no-repeat on the next tick, ledger actually written to disk, unwritable ledger, failed-post rollback, unconfigured key/room, and `main()` end to end asserting the room key never reaches stdout or stderr.
- **Toothless tests fixed**: the 401 test now asserts the dedicated branch's wording and that a 500 says nothing about tokens; the real-refusal test uses a port that was listening a moment ago, so it is Node's own error, and `fetchFailureReason` now walks the `cause` chain (and `AggregateError.errors`) so a user sees `connect ECONNREFUSED 127.0.0.1:53124` rather than `fetch failed`; the ledger-key test hard-codes `'local a1'` instead of calling `itemKey()`.
- **The NUL is gone.** `KEY_SEPARATOR` is an exported, documented `' '`. Sorry about the hour: the source file itself had ended up with a literal NUL byte in it, which is why `grep` went quiet on it.
- Duplicate blocked roster rows count once (dedup moved before the blocked branches); `options` dropped from `--json`; a flag with no value is an error rather than a silent default; `iak-pending` added to `package.json` `bin`; `examples/pending-launchd.plist` added as the timer sample, with a note to run it once without `--room` first.
- Left alone as agreed: the tailnet/token behaviour in `isTrustedGateHost`, and the secrets handling.

### Verification

Exit codes read from `$?` on the next line, never through a pipe.

```
PENDING_TEST_EXIT=0    ℹ tests 61   ℹ pass 61   ℹ fail 0
NPM_TEST_EXIT=0        ℹ tests 478  ℹ pass 478  ℹ fail 0
```

### Live run against the real fleet, `--json`, no `--room`

The peer's roster entry is a stable name now, so both hosts answered:

```
LIVE_JSON_EXIT=10
{
  "host_count": 2,
  "reached_count": 2,
  "pending_total": 5,
  "pending_below_threshold": 0,
  "pending": [
    { "host": "@peer", "id": "6125a374", "age": "10h 18m", "prompt": "Approve Bash command ..." },
    { "host": "@peer", "id": "19485d1f", "age": "9h 25m",  "prompt": "Approve Bash command ..." },
    { "host": "@peer", "id": "b2b47acb", "age": "9h 23m",  "prompt": "Approve Bash command ..." },
    { "host": "@peer", "id": "fa2bda32", "age": "4h 45m",  "prompt": "Approve Bash command ..." },
    { "host": "local",  "id": "cd703340", "age": "1h 0m",   "prompt": "Which model should ..." }
  ],
  "unreachable": [],
  "exit_code": 10
}
```

(Handles and prompts trimmed here; the tool prints them in full.)

And the honest-exit-code fix, live:

```
LIVE_THRESHOLD_EXIT=10
Nothing older than 1d 0h, but 5 items are pending below that threshold. Run without --older-than to see them.
```

Before this commit that same run exited `0`. Five approvals were waiting, the oldest of them for ten hours.

